PR #26, opened 2018-03-29
fix libretro buildbot msvc compile

markwkidd:
Twinaphex contacted me today to say that the XML generator code is causing their MSVC buildbot recipes to fail. Here is an example log with the error: http://p.0bl.net/178416

This PR was written on the run but it's working correctly when compiled for Windows x64 in mysys2 and I wanted to try merging it to find out if it will run through the MSVC buildbot in this form.

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -1,1100 +1,415 @@
-#include <ctype.h>
+#ifndef OSDEPEND_H
+#define OSDEPEND_H
 
-#include "driver.h"
-#include "sound/samples.h"
-#include "info.h"
-#include "hash.h"
-#include "datafile.h"
-#include "fileio.h"
+#include "osd_cpu.h"
+#include "inptport.h"
 
-/* Format */
-#define SELECT(a,b) (OUTPUT_XML ? (b) : (a))
+#ifdef __cplusplus
+extern "C" {
+#endif
 
-/* MESS/MAME configuration */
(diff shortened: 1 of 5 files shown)


grant2258:
-void print_mame_xml(FILE* out, const struct GameDriver* games[]);
+void print_mame_xml(const struct GameDriver* games[]);

thats causing the error the function takes a file handle

grant2258:
mark tree f42573f  Update info.c look looks like it has the wrong file

markwkidd:
I knew I shouldn't have tried to work on this in the middle of the night.
Oops!

I think everything is building OK now, except it takes the buildbot a while
to work so I still don't know if this fixed microsoft compilation. Finges
crossed.

grant2258:
is this code still valid or is it updated mark ?

markwkidd:
I made an emergency commit after this PR that has the code working again so you'll need to update your source.

Do you mind if I ask what your development environment is like? You probably need to `make clean` then `git pull` then `make -j4`

grant2258:
Im just using make on linux nothing fancy. It's compiling fine. If there is
more problems post updated links to the build errors ill try help out if I
can.

markwkidd:
Thank you, I appreciate it. I think that between the three of us who are
contributing we actually cover a lot of platforms just on our own testing.
Nice.

PR #53, opened 2018-04-13
fix Machine memory init and de-messify mame.c

markwkidd:


grant2258:
Mark ill move this code out of mame.c its not the best way to implement it. Ill do it like I done on mame2003 its the same thing done a different place.

PR #77, opened 2018-04-16
WIP silence scsp.c sound warnings

markwkidd:
These warnings seem more substantial than most so I'm more concerned about fixing them without knowing why things are the way they are.

Especially with the array size of `2`, which can't be right in the original source!


grant2258:
I would leave that as it is could have regressions just add flag to compile to surpress the warning until we find out for sure or keep a not of this one

markwkidd:
Yeah I think I'll not merge this for now till the implications are better understood

markwkidd:
Closing this as a PR and opening as an Issue

PR #101, opened 2018-04-18
Revert "fix rotation flipping still needs looked at"

grant2258:
Reverts libretro/mame2003-plus-libretro#98


grant2258:
not a good enough fix

PR #116, opened 2018-04-20
determine bios region from core options

markwkidd:
The one downside to this approach that I know of that it may require tweaking the 'internal' bios names for other drivers in order to make the region names standard like I have done here between `neogeo` and `stv`

The upside is that we can support more languages, etc.

Edit: I also rearranged the core options into my arbitrary idea of a better order, based on which settings people might go looking for first

### DIFF
--- a/src/drivers/neogeo.c
+++ b/src/drivers/neogeo.c
@@ -1454,13 +1454,13 @@ aof3
  ****/
 
 SYSTEM_BIOS_START( neogeo )
-	SYSTEM_BIOS_ADD( 0, "euro",       "Europe MVS (Ver. 2)" )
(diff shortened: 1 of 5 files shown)


grant2258:
neogeo has no taiwan option I wonder what the default behavior for that is

markwkidd:
From my `diehard` BIOS work I believe it will ignore a NULL or an invalid
bios name. Lord, now I'm worried that I remember that wrong

For this approach to work, drivers will definitely need to ignore invalid
BIOS names...

grant2258:
i think that probably is the default behavior as far as i can remember its does a string search for an index

PR #140, opened 2018-04-25
comment out brightness for now

markwkidd:


grant2258:
@markwkidd if we want to truely let libretro take over src/mame.c and all its related function will have to go and be re written. If we hack away at this core more and more it will get confusing what we are doing. Mame osd functions arent even being used at all. The input system is hacked in drivers and in the core im a bit confused now what is the goal here ?

PR #145, opened 2018-04-25
[hold] Initial attempt at re-map. Should be SNES layout for SF2 - needs test…

dankcushions:
…ing before merge. see https://github.com/libretro/mame2003-plus-libretro/issues/143


Makefile
```diff
@@ -2,7 +2,7 @@ DEBUG=0
 DEBUGGER=0
 SPLIT_UP_LINK=0
 CORE_DIR := src
-TARGET_NAME := mame2003_plus
+TARGET_NAME := mame2003-plus
```
retro-wertz: renamed again?

grant2258: nope was done be @dankcushions  so he could build this wont be changing

markwkidd:
I merged this just now because there is an opportunity to set up a Lakka test build that includes mame2003-plus, something that SapphireTactician has asked about. With this PR merged I can point that testing Lakka build to this commit

markwkidd:
ahh, I think there is some major problem with this code. it doesn't compile here and I see some other issues.

does this code compile for you @dankcushions ?

PR #156, opened 2018-04-28
Added retropad_layout option for common controller layouts.

dankcushions:
Added retropad_layout option, allowing for Modern (SFIV), SNES, and 'MAME Classic' (current MAME's Xbox 360 controller defaults) pad layouts. Defaults to Modern.

Version 2 of https://github.com/libretro/mame2003-plus-libretro/pull/155 - now works properly with dual stick shooters (Robotron, etc), and updates the MAME menu button labels after the layout is changed.


dankcushions:
I'd love to squash all the temporary retropie makefile changes but I get a bit lost doing rebases if they're not all in a row :/ If anyone else can do that, please do :)

markwkidd:
Squashed and merged! I think!

markwkidd:
Once your improvements have a chance to land, I'm kind of interested in refactoring the two input options to be something like:

retropad_mode: `modern|mame|snes|disabled`
keyboard_mode: `disabled|enabled`

I think that achieves the same functionality as `retropad|mame_keyboard|simultaneous` plus `modern|mame|snes|` but does it in a more intuitive way. Something to think about?

PR #158, opened 2018-04-28
[WIP] remove obsolete rstick to buttons core option

markwkidd:
Per this conversation: https://github.com/libretro/mame2003-plus-libretro/issues/157#issuecomment-385205111

@dankcushions is this what you have in mind? I wanted to get your thoughts before merging.


dankcushions:
I think not quite - this will just always make the right stick do the buttons. this works but it means we can't use the right stick for games that support a second analog axis (eg outrun accelerate). I think it would be better to map it to P1_JOYSTICKRIGHT_UP etc. we'd have to add extra members to the retrojsstate struct i guess.

markwkidd:
I've frankly never used this functionality. I'll either be patient or learn this code even better. I'm starting to figure it out as I follow your work.

PR #172, opened 2018-05-06
cleanups for rom loading code - no functional change

markwkidd:
Hi @grant2258 I think it's good to leave your null check for the driver name in place, just for now. A little extra checking would not hurt!

So my proposal in this PR is to clean up the comments, update `file_path.c`, and leave the logic as it is for our long-term use.

If that makes sense to you, feel free to merge this on in or give me the word and I'll so so. Thanks!

### DIFF
--- a/src/libretro-common/file/file_path.c
+++ b/src/libretro-common/file/file_path.c
@@ -383,15 +383,16 @@ const char *path_get_extension(const char *path)
  * Only '.'s after the last slash are considered.
  *
  * Returns: path with the extension part removed.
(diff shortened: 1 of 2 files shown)


markwkidd:
I spoke too soon. This fix to file_path may get reversed or changed somehow upstream because maybe some other software was relying on the existing broken/badly documented function. It's been discussed intensively on discord now.

grant's code works now. once things are clearer with libretro-common/file_path I'll resubmit this cleanup PR.

PR #192, opened 2018-05-12
[WIP] check extension - only .zip for romsets

markwkidd:
Howdy @grant2258 !

As you said, it's good to have checks on all of these strings for the various oddities people might throw at the load function.

I am proposing a check for the .zip extension because I believe, no matter what else we might be flexible on in the future, this core is only going to load ZIP files as its romsets.

Would you like to test this with your "no extension" emulationstation launch command line before someone merges it?

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -587,75 +587,79 @@ void retro_describe_buttons(void)
 
 bool retro_load_game(const struct retro_game_info *game)
 {
-    char            *driver_lookup;
-    int             orientation;
-    unsigned        rotateMode;
-    static const int uiModes[] = {ROT0, ROT90, ROT180, ROT270};
+  char            *driver_lookup;
+  int             orientation;
+  unsigned        rotateMode;
+  static const int uiModes[] = {ROT0, ROT90, ROT180, ROT270};
 
-    retro_describe_buttons();
-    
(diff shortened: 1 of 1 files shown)


dankcushions:
doesn't the api already dictate what you can load via https://github.com/libretro/mame2003-plus-libretro/blob/9c09d4199ec4f9520df1c3e4019d7a0f008d7607/src/mame2003/mame2003.c#L481 ?

also PLEASE PLEASE PLEASE stop putting refactors in these PRs :) at least have them in a seperate commit! so difficult to see the actual change!!! :)

markwkidd:
Evidently the API does not guarantee the ZIP extension. This all started
when someone had a odd emulationstation config with no file extensions and
it was segfaulting mame2003

And, the whitespace, why oh why was the function double indented, I
couldn't help it

dankcushions:
sounds like a retroarch issue, then. valid_extensions should do what it says! perhaps it doesn't do the same checks when loaded via command line? still a retroarch bug.

markwkidd:
Agreed. I will start thinking of this as a retroarch issue report that needs to be written up, along with the "rewind" issue. 

I think it's good to have workarounds here locally in the meantime.

grant2258:
just mere it anytime mark ill let you know or do a fix if there is any issues mate

markwkidd:
thanks grant, will do

PR #200, opened 2018-05-15
remove vestigal "Reset" and "OSD" mapping options in MAME menu

markwkidd:
I wanted to submit this as a PR so that is easier to track. 

I thought I removed the Reset and OSD mapping options a while back, since they already don't do anything. Maybe that was in mame2003. :man_shrugging: 

That being said this is a straightforward change that I have tested here in mame_keyboard and retropad modes just to make sure it doesn't knock off any of our other mapping code. I'll merge it now as well.


grant2258:
be careful when changing the size of this structure and changing the order youll mess with peoples cfg and they wont work as expected it would be better to set them as none people will have to remap again when doing these kind of changes

PR #219, opened 2018-05-20
split copyright notice and warnings back into two options

markwkidd:
I'm not against having two different core options for this but I'm just trying to keep the number somewhat manageable.

Does this 'four-way' selector make sense to folks?

According to the license we can't remove the copyright warning or distribute compiled versions with the copyright notice disabled by default. Any of our approaches is consistent with that requirement, I just wanted to mention it while we're looking at this part of the code.

### DIFF
--- a/src/mame.h
+++ b/src/mame.h
@@ -164,11 +164,6 @@ struct RunningMachine
 #define RETROPAD_MODERN			1
 #define RETROPAD_SNES			2
 
-#define NO_NOTICES        0
-#define ALL_NOTICES       1
-#define COPYRIGHT_NOTICE  2
-#define WARNING_NOTICES   3
-
 /* The host platform should fill these fields with the preferences specified in the GUI */
 /* or on the commandline. */
 struct GameOptions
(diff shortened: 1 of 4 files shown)


grant2258:
seems sensible to do it this way instead of 2 options. I dont display any of them warning anyway but it gives people the same choice in one  option. The retropie end might not like this because it has this set on the install script.

 # Set core options
    setRetroArchCoreOption "${dir_name}-skip_disclaimer" "enabled"
    setRetroArchCoreOption "${dir_name}-dcs-speedhack" "enabled"
    setRetroArchCoreOption "${dir_name}-samples" "enabled"

markwkidd:
I think I should return to the two different core options. It's simple to put that aspect back in place.

On one hand I do worry about too many core options, on the other hand life is short and Final Burn Alpha still has more than we do. Let me work on a follow up commit.

markwkidd:
I have pushed a commit that I think has us back to the two traditional core options for this, I'll run through the various possible combos before merging

markwkidd:
I cleaned it up a bit more and tested every combo. Seems to be working fine now and compatible with RetroPie-Setup.

PR #226, opened 2018-05-22
add support for unibios 3.3

markwkidd:
After the previous work, adding this BIOS was a snap.

I have only booted into the Neo Bomberman attract mode so far but at first glance Unibios 3.3 is working as expected.

@Wilstorm if we are experimenting with Unibios, I figure we might as well have this latest version supported by MAME.


markwkidd:
PS Unibios can be linked to within the terms of their license, so you can grab the appropriate file at http://unibios.free.fr/download.html

markwkidd:
screenshot 2018-05-22 17 11 44

Wilstorm:
@markwkidd - Just download from source and DAT will have the changes for the new BIOS?

markwkidd:
I have asked grant to update the dat because something seems fishy with dat
generation on my platform.

The code is all in place, I just am worried about the way my dat looked
once it was committed.

markwkidd:
Ok I should have waited 10 more seconds. Grant just added the DAT so you're
good to go!

PR #230, opened 2018-05-24
preliminary support for content-specific core options

markwkidd:
I have been gradually refactoring in this direction for a while now, and I believe now things are finally set up to selectively display core options based on the content that's being loaded.

This PR demonstrates how the logic works by only displaying the Neo Geo BIOS option when Neo Geo content is loaded, and the same for the ST-V BIOS selection.

The possibilities are quite broad from here. For example I would love to only offer the vector core options when a vector game is loaded.

I have done some basic tests and this seems to be working fine, but it is a pretty big change in terms of lines of code. And it is pretty late here. So I'm submitting this is a PR! :1234:


markwkidd:
I'm a little more alert now and thus I feel more qualified to merge this PR

The code to detect which games/systems are running is not elegant but that can be refactored into some better condition as these functions are used a few times and I (we?) figure out what will work best. 

The nvram bootstrap functions have a similar evolutionary trajectory, and since they're all about detecting which game is loading and taking action, maybe the bootstrap detection could even be merged with this eventually.

PR #232, opened 2018-05-24
[no functional change] demonstrate game-specific input labels

markwkidd:
I am trying to get to where I am ready to test use the code that @fr500 provided for joypad device subtypes. That's not what you're looking at here though.

Along the way it has now become much more straightforward to add game-specific input labels, starting with the Street Fighter 2 labels that fr500 incorporated into the diff patch.

**With this PR, Neo Geo games and SF2 clones will display game-apporpriate names along with the MAME button numbers. Instead of just `Button 1` in a Neo Geo game, you will see `BTN1: Neo Geo A`. For Street Fighter 2, you will see the button number and then things like `Weak Kick`, `Strong Punch`, etc.**

## This is not meant to make any functional change to way inputs work now

I've taken care to only refactor and enhance the labels. I am submitting this is as a PR because the input system is important, and even though I have been careful I wouldn't mind another pair of eyes (@dankcushions ?)

Once this piece is finalized it does also serve as a building block towards the input subtypes!

### DIFF
--- a/src/mame.h
+++ b/src/mame.h
@@ -160,9 +160,6 @@ struct RunningMachine
 #define ARTWORK_USE_BACKDROPS	0x01
 #define ARTWORK_USE_OVERLAYS	0x02
 #define ARTWORK_USE_BEZELS		0x04
-#define RETROPAD_MAME			0
-#define RETROPAD_MODERN			1
-#define RETROPAD_SNES			2
 
 /* The host platform should fill these fields with the preferences specified in the GUI */
 /* or on the commandline. */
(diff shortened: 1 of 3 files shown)


markwkidd:
I feel like I need to say that I'm aware this is a very simple implementation, but I think it can get more sophisticated and concise once we are farther along.

That's the direction I'm move towards with aspects of this implementation like the new `enum`s so that we can eventually use switches and loops instead of a bunch of assignment operators in a long line.

grant2258:
@markwkidd one thing to keep in mind retropie doesn't use ra default controller settings the user sets this up themselves they choose buttons for a b x y per controller. 

Keep this in mind when getting feedback. Hope it works out for whatever you trying to setup for xbox/snes type controllers

PR #250, opened 2018-05-31
bring back the sample core option for toggling alternate audio

markwkidd:
Annoyingly this core option returns from the grave. I suppose I have the means now to only display it for the games with the new alternate soundtracks


grant2258:
pretty sure you removed the sample check from code itself mark if i remember right

markwkidd:
I thought I did too but it looks like it's still there in common.c

grant2258:
ok roger that sorry just had memorys for the 2003 fixes never mind me mate!

PR #258, opened 2018-06-04
WIP: add core option to toggle use of mame remapper

markwkidd:
This for some reason isn't working yet but I wanted to put up a PR in case someone else is also working to make a core option to to toggle the availability of the MAME remapper. The default would be to have the MAME remapper disabled.

The thing I'm most interested in is using this core option to not only show or hide the remapper in the MAME menu but also to toggle the loading and saving of .cfg files.

This allows the user to keep some cfg files in place / not delete them but still avoid having them applied by MAME. That seemed like a good way to approach things.

I'll fix this soon...


markwkidd:
I'm re-implementing this over in the fake retropad subtype branch

PR #264, opened 2018-06-05
standardize content-specific flags and boilerplate for showing/hiding core options

markwkidd:
Hi everyone, and paricularly @grant2258 . I'm tagging you because this refactors your OST detection code a little bit. Now it will only display the core option for games that have the alaternate soundstracks as well.

I'm hoping that this commit yields a nice flexible way of handling content-specific functionality. I took the time to make some boilerplate, not activated yet, which will also show or hide the vector menu, dual joystick option, etc.

All the boilerplate still needs is some more tests for these attributes so their flags can be set.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -239,24 +239,6 @@ struct GameSamples *readsamples(const char **samplenames,const char *basename)
 	struct GameSamples *samples;
 	int skipfirst = 0;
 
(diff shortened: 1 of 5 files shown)


grant2258:
look good thanks for the heads up mark

PR #296, opened 2018-06-18
TESTERS NEEDED: implement dcs hack with Killer Instinct support

markwkidd:
I implemented the modifications to dcs.c that @arcadez posted here: https://github.com/libretro/mame2003-plus-libretro/issues/167

I have tested this with Killer Instinct on a 10-year old laptop which struggles to play Killer Instinct at full speed. With the new patch action feels much more fluid. Could anyone else give this branch a try?

### DIFF
--- a/src/sndhrdw/dcs.c
+++ b/src/sndhrdw/dcs.c
@@ -455,6 +455,15 @@ void dcs2_init(offs_t polling_offset)
 		for (i = 0; i < 0x400; i++)
 			dcs_expanded_rom[0x400 * page + i] = romsrc[BYTE_XOR_LE(0x1000 * page + i)];
 	
+    /* install the speedup handler */
+    if (options.activate_dcs_speedhack)
+    {
+      dcs_speedup1 = install_mem_write16_handler(dcs_cpunum, ADSP_DATA_ADDR_RANGE(0x04f8, 0x04f8), dcs_speedup1_w);
+      dcs_speedup2 = install_mem_write16_handler(dcs_cpunum, ADSP_DATA_ADDR_RANGE(0x063d, 0x063d), dcs_speedup2_w);
+	    dcs_speedup3 = install_mem_write16_handler(dcs_cpunum, ADSP_DATA_ADDR_RANGE(0x063a, 0x063a), dcs_speedup3_w);
+      dcs_speedup4 = install_mem_write16_handler(dcs_cpunum, ADSP_DATA_ADDR_RANGE(0x0641, 0x0641), dcs_speedup4_w);
+     }
+	
 	/* create the timer */
 	dcs.reg_timer = timer_alloc(dcs_irq);
 	dcs.sport_timer = timer_alloc(sport0_irq);


markwkidd:
@arcadez I agree.

Do you know a list of all of the drivers that have the DCS speedhack option? I've looked at `dcs.c` and I'm having a hard time figuring out which games are actually affected

arcadez:
@markwkidd There is another speedup in the dcs2_init but i dont know if it's turned on
by default, certainly i see no way to flip it off or on unlike the dcs speedups we already use

it might be worth disabling the polling check..?? as i assume then it would be active..??
	/* install the speedup handler */
	/*if (polling_offset) */
		dcs_polling_base = install_mem_read16_handler(dcs_cpunum, ADSP_DATA_ADDR_RANGE(polling_offset, polling_offset), dcs_polling_r);

arcadez:
Mortal Kombat II in midtunit.c 
all games in midvunit.c
Crusi'n USA games use DCS in midvunit.c
War Gods use DCS 2 in midvunit.c
Killer Instinct games use DCS 2 in kinst.c

i'll need to check one other varient of the dcs to find the other games

arcadez:
@markwkidd Here's a handy hint, if you wanna know for example which drivers use a
sound core you only need to do a search of your source or the git and anything that
uses it will pop up.............

MACHINE_DRIVER_START( dcs_audio )
	MDRV_CPU_ADD_TAG("dcs", ADSP2105, 10000000)
	MDRV_CPU_FLAGS(CPU_AUDIO_CPU)
	MDRV_CPU_MEMORY(dcs_readmem,dcs_writemem)

	MDRV_SOUND_ADD(CUSTOM, dcs_custom_interface)
MACHINE_DRIVER_END

In the case of DCS all the drivers hook to the above via "dcs_audio" flag

markwkidd:
Ah that makes sense. I have not queried based on sound core before. Thanks!

Darknior:
Really excellent, good work for Killer Instinct :)
I have compile the last source code to try it yesterday but it doesn't work for me.
What set of ROM must we use ?
Must we use also CHD files ?
Thanks

6alileo:
Hey guys - does this work on ARM? CHD file?

markwkidd:
Yes it does work on ARM. Yes it does need the correct CHD file (mame2003-plus takes the same romset files and CHD version as mame 0.78/mame2003 for these games)

6alileo:
@markwkidd Cool - got both to load on the XU4. Both have choppy sound except on Ki1 indoor stages work a bit better. Wonder if there's some compiling flags I can use for armv7

6alileo:
I was able to compile with either and sound is still choppy. Indoor stages
get smoother. Maybe the XU4 doesn't have enough grunt. Is this core using
any kind of threading so audio is on a separate core?

grant2258:
I think we can reach a point and say you need a faster processor. you cant do the audio on another thread it uses multi processors and interrupts for other processors for audio its not as simple as threading the audio to another thread  like you would when writing a game for a pc

6alileo:
No problem. Thank you for getting this done. It's a pretty big deal! I will
try on my PC

PR #300, opened 2018-07-03
Null the string to avoid the random text box

grant2258:
This should fix the random text box problem that appears sometimes


markwkidd:
Makes sense! Merging by phone

PR #323, opened 2018-07-31
Fix flac

grant2258:
I have tested this seems ok in all ive tested.

If this doesnt merge in the main branch ill ust do a normal pull req to master dont wrestle with it make is only a few files can easly be done my end it it doesnt play nice. just let me know if there is any issues putting in the master branch


grant2258:
@markwkidd i cant help bu think int c_sample; should also be changed to int c_sample=0; that might be enough in itself to stop this behavior but its best to catch all ill add the int c_sample=0; in another pull req sorry just finished a shift at work and am blurry eyed

PR #326, opened 2018-07-31
Potential graphical fix for Rev-X

arcadez:
- 9th May 2007: Aaron Giles - Having recently touched all of the TMS34010-based games led me to start looking into some long-standing issues with some of the Williams/Midway Z/Y/X/Wolf-unit games. The first thing to do was to add save state support to them, because playing Revolution X more than once to get to the place where the video craps out is not for the faint of heart. This was relatively straightforward, and also had the side benefit of making some useful data available in the debugger (such as the local_videoram array and the dma_registers array). Side note: any array or pointer you register with state_save_register_global_array/_pointer automatically becomes a viewable option in a memory window in the debugger. This is very handy if you need to expose any internal state. Anyway, it turned out that the Revolution X bug was caused by over-aggressive masking in the blitter code. It was masking out one too many bits in the clipping window and this led to the right edge of the clip window being less than the left edge, effectively clipping out everything during the blit. Hence the black screen.

- 0.114u2: Aaron Giles fixed Revolution X gameplay/finish a level.

### DIFF
--- a/src/vidhrdw/midtunit_vidhrdw.c
+++ b/src/vidhrdw/midtunit_vidhrdw.c
@@ -751,8 +751,8 @@ WRITE16_HANDLER( midtunit_dma_w )
 	/* clip the clippers */
 	dma_state.topclip = dma_register[DMA_TOPCLIP] & 0x1ff;
 	dma_state.botclip = dma_register[DMA_BOTCLIP] & 0x1ff;
-	dma_state.leftclip = dma_register[DMA_LEFTCLIP] & 0x1ff;
-	dma_state.rightclip = dma_register[DMA_RIGHTCLIP] & 0x1ff;
+	dma_state.leftclip = dma_register[DMA_LEFTCLIP] & 0x3ff;
+	dma_state.rightclip = dma_register[DMA_RIGHTCLIP] & 0x3ff;
 
 	/* determine the offset */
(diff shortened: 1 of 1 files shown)


arcadez:
Could be dangerous this one as all the Midway T and W unit games use this code
i think this will be the fix that was done for Rev-X but the above games would need
testing to make sure there is no breakage

grant2258:
could it be wrapped in an if condition with the machine driver name ?

grant2258:
thats why i was thinking a least it contains it  to one game for now

arcadez:
Maybe i shouldn't be so negative of thought im sure the video change will be fine for all
the games, i'll leave the fix here for now as im not sure all that many would benefit from
it anyway due to the calibration setup which will be required for rev-x.

Wilstorm:
@arcadez - I'm not sure which games are the T and W unit games but if you need me test after a tweak just let me know.

arcadez:
@Wilstorm here's a list of the games........

		* Mortal Kombat (T-unit version)
		* Mortal Kombat 2
		* NBA Jam
		* NBA Jam Tournament Edition
		* Judge Dredd (prototype)
		* Mortal Kombat 3
		* Ultimate Mortal Kombat 3
		* NBA Hangtime
		* NBA Maximum Hangtime
		* 2 On 2 Open Ice Challenge
		* WWF Wrestlemania
		* Rampage World Tour

PR #327, opened 2018-08-01
Samples enhancements

grant2258:
is ready to bring in if it works


grant2258:
guessed as much the usual github magic

PR #341, opened 2018-08-27
send driver log to debug

grant2258:
you can use this to clean the log spamming up if you like just take note your debug log will still be spammed but it is debug information its where it should be.


markwkidd:
Wow, nice, thanks

PR #361, opened 2018-09-02
Cave PGM WIP

arcadez:
Dont merge but feel free to test if ya's wanna ;)

### DIFF
--- a/src/drivers/pgm.c
+++ b/src/drivers/pgm.c
@@ -239,7 +239,11 @@ void pgm_dw2_decrypt(void);
 void pgm_djlzz_decrypt(void);
 void pgm_killbld_decrypt(void);
 void pgm_pstar_decrypt(void);
+void pgm_ket_decrypt(void);
+void pgm_espgal_decrypt(void);
+void pgm_py2k2_decrypt(void);
 
+READ16_HANDLER( pgm_calendar_r );
 READ16_HANDLER( pgm_asic3_r );
 WRITE16_HANDLER( pgm_asic3_w );
 WRITE16_HANDLER( pgm_asic3_reg_w );
@@ -416,6 +420,17 @@ WRITE16_HANDLER( pgm_calendar_w )
 	}
 }
 
+/*************************************
+ *
+ *	NVRAM
+ *
+ *************************************/
+
+static READ16_HANDLER( nvram_r ) /* ddp3blk needs this to boot */
+{
+	return ((data16_t *)generic_nvram)[offset] | 0xfff0;
+}
+
 /*** Memory Maps *************************************************************/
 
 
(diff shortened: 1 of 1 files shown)


arcadez:
So i've lucked out on a less than busy night while browsing the different MAME sources on here
a pgm hybrid driver someone ported to AdvanceMAME from another project

https://github.com/amadvance/advancemame/blob/master/src/drivers/cavepgm.c

There was alotta incompatable code in there ARM7 CPU not withstanding so i stripped it
all out just leaving what was needed dumped it into my PGM driver and reworked it to the 
MAME84 codebase, and bingo the games all booted and play just shy of full speed.

The sound is quite low but nothing the vol button on my TV wouldn't sort but i mind
the FBA boys menting there is a better ICS2115 sound core available which will greatly
improve the music and sfx in the PGM games whether or not it can be backported to
ole MAME as always will be the question.

Anyway i've now backported the code once more to MAME78 feel free to test the games
out if ya wanna they should be working but again with this ole core ya never know :)

For testing purposes
https://www.sendspace.com/file/faqpi3

arcadez:
I was holding out for a ICS2515 sound core with the improvements but it's not going to be
compatable with this ole core so time to merge this, just mind and turn up the vol and
you'll be fine i think :)

grant2258:
youve been busy this weekend mr arcadez ill test in the when i get up hitting the hay. I was fighting with github it nearly won

arcadez:
Yeah i seen the driver and i thought i'll have that!! as im partial to those games and on the xbox
we only had FBA to play em but the speed drops a fair bit here and there, in the older MAME core
and keeping the older PGM video code they're super fast now, and after my original port work
it was a breeze to then drop it in here also.

As is often the case though on the ole box you gain on one hand but lose on the other as the sound is 
annoyingly low in ole MAME vs FBA not that a vol knob wont sort it though ha ha

grant2258:
lol well the v60 core is updated as well so thats sitting at mame 122 version it should just prort right into 84 as well if you only take the core files

arcadez:
I was curious as to why you updated it..?? anyway if it now has the trap and stack fixes from
MAME97 we could have two more SSV games Eagle Shot Golf and the rather tasty
Mobil Shot Gundam Final Shooting i have all the code for them it would be easy enough
i think to drop it in here.

Just a heads up though they changed the clock divders for the V60 in MAME97 so when i
updated the V60 in MAME84 i had to reclock all the drivers that used it as the games would
take a huge performance drop with the older clocks.

grant2258:
I originally updated this because of the sound breakage and ultrax not working. I think the sound now and in the old v60 is down to clock speeds the v60 is set to 

https://github.com/grant2258/mame2003-plus-libretro/blob/7a0e8adc345c1430bf35323685fa89fe9b6b321c/src/drivers/ssv.c#L2841

setting it to 16000000 makes the sound a lot more pleasant but its still not right.

grant2258:
if im going to have to work out clock speeds for every v60 game it might be worth reverting it back unless its an easy converstion

arcadez:
It's ok buddy in this old core it's only 3 drivers to change multi32.c, ssv.c and system32 i'll 
import the clock changes for ya.

grant2258:
I never applied the clock changes in the cpu exec

PR #372, opened 2018-09-10
fix es5506 sound in ssv games

grant2258:
after a lot of debugging and messing about in the wrong places i found the problem thank god


grant2258:
well this bug was alive since 2013 commit not many games use the 5506 and it really was a easy mistake to make to be fair all well that ends well. I was convinced it was the interrupts just shows you how wrong you can be

PR #374, opened 2018-09-11
update irem decryption tables to mame 174 fixes nbbatman 

grant2258:
the sound is very messed up at the first boss this fixes that im not 100% sure if the pitch is right or not but that really is a separate issue if it isnt right. These fixes might fix other games with sound issues as well so might be worth testing the m107,m90 and m92.

@arcadez if you want to add this to you core just replace all the array tables and the .h file

### DIFF
--- a/src/drivers/m107.c
+++ b/src/drivers/m107.c
@@ -684,7 +684,7 @@ static DRIVER_INIT( firebarr )
 	RAM = memory_region(REGION_CPU2);
 	memcpy(RAM+0xffff0,RAM+0x1fff0,0x10); /* Sound cpu Start vector */
 
-	irem_cpu_decrypt(1,firebarr_decryption_table);
+	irem_cpu_decrypt(1,rtypeleo_decryption_table);
 
 	m107_irq_vectorbase=0x20;
 	m107_spritesystem = 1;
--- a/src/drivers/m90.c
+++ b/src/drivers/m90.c
@@ -934,7 +934,7 @@ static DRIVER_INIT( riskchal )
 
 static DRIVER_INIT( shisen2 )
 {
-	irem_cpu_decrypt(0,shisen2_decryption_table);
+	irem_cpu_decrypt(0,matchit2_decryption_table);
 }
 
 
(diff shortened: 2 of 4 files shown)


arcadez:
It's the raster graphical effects which are the cause of the sound problems on the xbox at least
with fps enabled you can see it drop to as low as 11fps then back up to 45fps or so hence
the sound stuttering.

In ShmupMAME i see he actually disabled the raster irq triggers in the video code this keeps
the games at a high framerate but it does introduce some graphical niggles in proccess

arcadez:
BTW just for the record this came up before, long story short the boss music is different
in the World version vs the USA version, some users dont realise there was a new version
dumped which then became the parent.

So when they play the game across different versions of MAME using the parent rom
they will think the music is wrong on the bosses.

grant2258:
No the music is garbled just random high pith noises

grant2258:
here is the link of the fixed and broke samples just finished work. in theory this should fix any other broken irem sounds if its decryption related. 

https://github.com/grant2258/mamemagic2003/tree/master/nbbatman_samples

Wilstorm:
@grant2258 - That's nice, the difference is definitely noticeable! I think it's in your vein of music too but I went to Metallica's Hardwired Worldwide tour a couple a nights ago. They still can rock it old school for a bunch of old guys, I was impressed. Whew loud, I crawled into bed about 2 and was dragging ass the next day at work, I'm getting to old to skip half nights sleep.

grant2258:
nice metallica are always a good band live no matter what set they do! last concert i went to was avenged sevenfold about two years ago!

Wilstorm:
You posted the one video that was really good. I need to expand more European Metal. I grew up on Ozzy, Accept, Scorpions, Judas Priest, Saxon, Motörhead, Iron Maiden, Black Sabbath, etc. for overseas but don't know much beyond that. The last time I saw them was in '89 during the 'Justice For All' tour. They were so much more gracious and thankful to the audience at the end vs back then. I go to about every concert that comes except country. The stage show/layout was really well done. Good pyrotechnics.

PR #377, opened 2018-09-17
repair for insufficient checking against null

360CodeSafe:
Hi all,

There is a possible null pointer dereference found by Qihoo360 CodeSafe Team, see
If only 'entry' is checked, 'associateEntry' could sitll be a null pointer (if 'associate' is 0).
It is better to check both 'entry' and 'associateEntry' against null pointer.
Then there will be OK if 'associateEntry' is passed as argument to function 'AddActionWatch()' in which  the 2nd argument is dereferenced in line 9558.

### DIFF
--- a/src/cheat.c
+++ b/src/cheat.c
@@ -9507,11 +9507,10 @@ static void WatchCheatEntry(CheatEntry * entry, UINT8 associate)
 	UINT32		i;
 	CheatEntry	* associateEntry = NULL;
 
-	if(associate)
-		associateEntry = entry;
-
-	if(!entry || !associateEntry)
-		return;
+	if(!entry || !associate)
+    return;
+  
+  associateEntry = entry;
 
 	for(i = 0; i < entry->actionListLength; i++)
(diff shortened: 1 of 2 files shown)


grant2258:
I believe this function is done correctly as it is even if  associate is null associateEntry = entry is set to entry or null. if   entry is null the function exits regardless of associate vaule

watching static void WatchCheatEntry(CheatEntry * entry, UINT8 associate)
{
	UINT32		i;
	CheatEntry	* associateEntry = NULL;

	if(associate)
        associateEntry = entry;

markwkidd:
I think static analysis is handy so I'm willing to clean up a couple of flags in the hopes that future analysis will be that much more focused on issues with more significant consequences. Thank you @360CodeSafe and please feel free to keep a conversation open if you notice things in other areas of the codebase.

I think the fileio.c change is good as proposed. I would like to implement the change to cheats.c in a way that is a little cleaner and obvious to understand in my opinion.

Issue #380, opened 2018-09-28
4-way-restriction


robertvb83:
There is a software based attempt to change the behaviour of 8way joysticks @ 4-way games in xboxdrv https://github.com/RetroPie/RetroPie-Setup/wiki/Universal-Controller-Calibration-&-Mapping-Using-xboxdrv

Could something like this be included in mame2003-plus for all the 4-way games as an option?

There has been attempts in mame as discussed in retropie forum https://retropie.org.uk/forum/topic/11931/using-xboxdrv-for-4-way-restrictor-on-complete-rertropie-setup-already-using-dragonrise-usb-controller-setup/20

See especially last few posts

Wilstorm:
@robertvb83 - Thanks for the links! That's exactly what I was looking into today using Mediamogul's guide. I think this would be the cat's meow for Plus. I have the first part complete but looking at Spud's script I am not sure. I'll have to spend some time trying to figure out what I would need to do as I am really new to attempting 4/8-way game restriction with a controller.

That would be neat to only need $3 variable and simplify the whole process.

robertvb83:
Does anybody know how the algorithm in xboxdrv works?
Mame approach was like this: if joystick moves from up --> up/right then output is right. So on movement the unchanged direction is ignored.

To be fair this IS a prediction and there is a chance of false movement. Also if changing from up --> down/right it is hard to deal with the diagonal input. As it is not clear what the intendet move was, however i would make the algorithm go down in that case. Because it feels mote likely, why would i move so far if i just like to go right.

There have also been speeches against this implementation and i do not know what was the conclusion for mame finally or what it is like in current mame.
I think as long as its optional, this is a clear benefit. I have trouble playing games like pacman with my 8-way-joystick and would love to try the feeling of a software assisted 4-way predictor. Bomberman can be really horrible because the fucker simply wont move on any diagonal inputs

grant2258:
Without knowing what is happening in the game, and analyzing it,  or having more information available from the stick itself (i.e. more than 4 pieces of information),

There is no way to make a determination of what is the "probable intended" direction with an accuracy level higher than a "coin toss" it will be a per game bases.  As this is pretty much the accuracy one gets by default, the actual net gain is zero. Use the correct restrictor jobs done

robertvb83:
@grant2258 how come that there are software based solutions with people getting a satisfying result as can be read in the above linked retropie thread?
[...]

PR #384, opened 2018-10-02
fix up the logic for 4way

grant2258:
This has been tested on windows 64 with retroatch with a xbox360 and xbox one and an dragonrise zeroday encoder with a sanwa jlf. The same controllers where tested on retropie. I didnt want to commit this until I was 100% happy with it.

A little note on xbo360 controllers you if your using the analog as digital please put the dead zone up to at least 0.600 if your having wondering problems. These sticks more so when they age are notorious for is as well as ps3 ones.

There are 3 options for 4 way in the options its on original by default original is mames native code. I personally recommend last_pressed. 

@robertvb83  said he was having problems with bomberman. Its set to 8way in the mame source in the drivers and im having no problems with the game that way. (only tested on my sanwa stick to be fair). 
I have the source changes to make it 4way if someone can tell me how to reproduce the problem. I will change it.

hopefully this is the last update on this

### DIFF
--- a/src/inptport.c
+++ b/src/inptport.c
@@ -2181,11 +2181,10 @@ ScanJoysticks( struct InputPort *in )
 		}
 
 		/* Only update mJoy4Way if the joystick has moved. */
-		if( mJoyCurrent[i]!=mJoyPrevious[i] )
+		if( mJoyCurrent[i]!=mJoyPrevious[i] && !options.four_way_emulation)
 		{
 			mJoy4Way[i] = mJoyCurrent[i];
-			if (!options.four_way_emulation) //start use original code
-			{
+
 			  if( (mJoy4Way[i] & 0x3) && (mJoy4Way[i] & 0xc) )
 			  {
 			  	  /* If joystick is pointing at a diagonal, acknowledge that the player moved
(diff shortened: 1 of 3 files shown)


grant2258:
this is intentionally just set up for last pressed for now is seems the best logic to go with. Staying true to a 4way you keep  your last direction when your on a diagonal ie no switch pressed and your direction will change when you reach your final none diagonal direction on the joystick.

robertvb83:
for bomberman the control is weird. with my 8way sanwa it is like this: if i press right and change into up/right or down/right diagonals bomberman is changing direction to up or down (OK!)
but if going down and change into down/right nothing is happening keeps running down, whereas down/left will immediately turn left. Going from up in the right or left diagonal no direction change will occur.

i dont understand that bomberman is tagged as joy8way because it is obviously only up down left right, does not make sense to me at all

robertvb83:
is this right?
mJoy4Way[i] ^= last_direction

grant2258:
to be honest i really dont understand why you guys want the diagonal direction button to do something it doesnt exist on a 4way . I should just be ignore like it is done in last_pressed

choose last_pressed and it effectively emulatedsa 4 way joystick by completely ignoring  the diagonal button all together and keeping the last direction until another single switch is selected. This is effectively what a restriction does. Mame does not do this is pauses dkong when you hit a diagonal.


I agree xor is int the best fit but there is no way round the extreme diagonal if you choose to use diagonal directons 

mJoy4Way[i] &= ~last_direction is perfect for up to right or up  left but there is no solution for up/right to down/left you would need to guess a 50/50. or center to diagonal. Thats why the only sane choice is last pressed. 

test donkey kong / kung fu master and with last pressed forget xor or any other diagonal guess work.

SaucySarah:
It seems Bomberman wasn't actually changed to 4-way joystick until version 0.187 !

```
0.190: Soundlatch modernization for clone Dynablaster / Bomber Man (bootleg, set 1) [AJR].
**0.187: Use 4-way joysticks (as service manuals suggest) in Dynablaster and clones [AJR].**
0.184: Hammy and The Dumping Union added clone Dynablaster / Bomber Man (bootleg, set 3). Changed description of clones (bootleg) to 'Dynablaster / Bomber Man (bootleg, set 1)' and (bootleg, alt) to 'Dynablaster / Bomber Man (bootleg, set 2)'.
```

robertvb83:
Just play bomberman for snes with your 8way joystick. It is very frustrating getting bombed because bomberman did not move on diagonal.

You CAN NOT avoid going into the diagonal spot accidentially! It is physically different to a real 4way restrictor

Please grant try bomberman withyour 8way its hard to describe

Wilstorm:
@grant2258 - I have an appt. here but just doing a quick test things look better. You completely changed some things, not even the keyboard is working the same. There seems to be a definite difference between the settings Original vs. XOR or Last Pressed. Donkey Kong and Pac-man feel pretty good. With the 'Original' setting Pac-man may turn accidently on long straightaways but with XOR or Last Pressed it felt good. I am having a hard time telling the difference between XOR or Last Pressed to know which might be better in certain games. I will test Mr. Do and Kung-Fu Master later but have to get going here.

grant2258:
glad you feel its an improvement the use of xor was to cut any diagonals and and set bit masks again it didnt  work out the only realy winner here is last pressed for me. Xor wasnt fully implemented anyway there is no algo that can work out what yu want to to that will come from the natural joystick movement when you ignore the diagonal. last pressed basically translated to keep last direction until a new direction is reached.

 Think of original as keyboard based and new as joystick based. One doesnt compliment the other as they are two different mediums. That why the mame guys left it the wa

Wilstorm:
@grant2258 - I did test both Mr. Do and Kung-Fu Master last night for about a half hour. I spend time playing 'Original' to get a feel and then with the new modes. I wiggle the joystick a little on some games to see if it's easy to tun on straightaways and what not.

Both worked well. Mr. Do with 'XOR' had that same lag like Pac-Man. I don't know if it hits a diagonal dead spot or what happens here when turning but it has a very brief millisecond of outage/pause that you can feel when you play. It's not long enough to get you killed or anything but it does have a slightly different feel to it. I do prefer 'Last Pressed' here for sure.

Kung-Fu master seemed less noticeable to me. Flipping left/right constantly to kick the enemies off the floor felt smooth with either but overall I am leaning more toward 'Last Pressed'.

I will keep testing 4-way games and let you know if I come across anything strange. If you need me to test any other games just let me know.

grant2258:
well there has been updates since its only last pressed now. conversation is best discussed in issues because you end up with multiple conversations when the code has changed in the pull req threads. Please see here mr  @Wilstorm   https://github.com/libretro/mame2003-plus-libretro/issues/380#issuecomment-426611384

PR #385, opened 2018-10-02
Add another 42 hacks of wofch.

SumavisionQ5:
Thanks to arcadez for adding the driver for wofch and fix  for player 3 input, so that I can add more wofch hacks!
Please use latest HBMAME ROMsets. 
Feel free to delete the drivers that you don't like. 
There are still a few drivers of wofch hacks left. I will try to add them after my test. 
Rgds.


arcadez:
Aye i guess you like that game :)

SumavisionQ5:
@arcadez, thank you. I'll add them to the what's new later.

SaucySarah:
42 Hacks of 1 game ??!!

Wilstorm:
@SaucySarah - Come on Sarah, I'm thinking replay value here, why have 2 or 3 hacks, when you can have FIFTY--that's right, the big 5-0, and more to come it looks like! :)

arcadez:
I kinda agree with @SaucySarah 50 versions of the same game FFS, sometimes less is defo more :)

PR #386, opened 2018-10-02
update bombrman to 4way and the final changes to the new 4way code

grant2258:
bomberman changed to 4 way joysticks and final 4way joystick update unless someone comes up with a decent working algo for extreme diagonals this is what we have.

please test with 
Dkong 
hungfu master 
and mr do

there is two four way options original and new you will need new to play dkong as the old version locks up with diagional. I pretty much prefare this new version


grant2258:
hoing to work will check in when i finish my shift

PR #390, opened 2018-10-04
Add another 15 hacks.

SumavisionQ5:
Add last 5 hacks of wofch and 10 best hacks of wof.
There are hundreds of hacks for Warriors of Fate(AKA Tenchi wo Kurau II or Sangokushi II) in HBMAME 0.202. I tried my best to pick the best ones and ported them to MAME 2003 Plus. 
If you like Capcom's beat'em up games, it would be a pity to skip the hacks I choose.
For your information, wofch is an enhanced version of wof released by Capcom. 
Wofch is released for CPS CHANGER, with new power-up special for each character, which brings in fantastic combos and greatly increases the fun.
It was later ported to PlayStation and Sega Saturn, and was published in Japan only.
For PS version, There is a cheat to enable hidden mode.
At the title screen, press R1, Triangle, L1, Square, Cross, Circle, R1. 
The title will change from red to blue if successful. 
Then you can input down, up, down, up + attack to do the power up special with 4 of the warriors except the big guy(Zhang Fei) in yellow clothes. 
To do these four characters' most powerful combo, get close to an enemy, grab it and punch it twice, jump and input attack, while you drop from sky, quickly input down, up, down, up + attack.
To let out Zhang Fei's power up special, you have to get close to enemy, and input forward, back, down, up + attack very quickly.
For CPS CHANGER version, press Pause and Start button for more than 1 second to enter Option Menu. 
Then press: R-X-L-Y-B-A-R to enable hidden mode. Both game title and blood color would become blue.
Of course, you can use the cheat code I provided for MAME2003 plus to enable hidden mode directly.


SumavisionQ5:
I will update CHANGELOG.md later. That's a long list now :)

PR #392, opened 2018-10-04
Revert "Add another 42 hacks of wofch."

arcadez:
Reverts libretro/mame2003-plus-libretro#385


arcadez:
Your meant to say please ;) however i can do ya kofth i have the code already in another project
i guess it could be ported across sometime should i fell like it.

Issue #376, opened 2018-09-17
A possible NPD?


360CodeSafe:
Hi all,

There is another possible null pointer dereference found by Qihoo360 CodeSafe Team, see

pointer is check against null here:
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L438
but does nothing even if 'src2val' is a null pointer.

then the function 'check_irqs()' is called;
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L443
which is definited here:
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L388
in which function 'generate_exception()' is called here:
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L392
which is definited here:
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L367

in function 'generate_exception()', pointer 'src2val' is used without checking against null, which may be a possible NPD.
https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L372

Can anyone who is familiar with this project can help me to confirm this?

Cheers!
Qihoo360 CodeSafe Team.

grant2258:
as far as i can see from a quick glance the init routine sets the info so this shoudnt ever happen

markwkidd:
I agree with grant that it seems like the only way that src2val would be if its `malloc` failed in the init function here: https://github.com/libretro/mame2003-plus-libretro/blob/92e4d7b69ce597df61a2eee7e32fd60d12e9d6f9/src/cpu/asap/asap.c#L481

However I don't see any reason not to make the existing null checks a little more robust for that pointer -- I may put together a PR

markwkidd:
@360CodeSafe I have merged a patch and I will close this issue for now. Please feel free to reopen it or post again if there are any further developments

PR #395, opened 2018-10-05
check for a null pointer a few places in asap cpu

markwkidd:
the point of this PR would be to eliminate some static analysis flags raised by @360CodeSafe in this issue: https://github.com/libretro/mame2003-plus-libretro/issues/376

I don't think adding these checks actually has any effect but maybe it would help future static analysis be more focused on issues more likely to cause problems.


grant2258:
Problem with static analysis is you do get false positives. Sometimes a pointer is set to null intentionally elsewhere In  the code to skip a set condition you really need to follow the whole loop of the code before making assumptions

PR #404, opened 2018-10-18
update vsnes soundcore bugs and fix a log_cb error

grant2258:
This should fix the volume levels up there is a significant drop in the monitor levels in audacity the volume might actually be to low now but leave things as are until its tested on the switch. Please try playchoice 10 games as well make sure they are ok as  well as they use this sound core as well


grant2258:
volumes

PR #406, opened 2018-10-19
WIP port nayslayer's vector fix from mame2003

markwkidd:
I tried to port @nayslayer's patch over.

Mainly the differences had to do with changes in the core option rigging between the two cores, although there were also some variables in vector.c that no longer matched between the two cores. At any rate this should do the trick.

I want to work with this a bit more to make sure that I've got these core options working and updating as intended. To be continued.

### DIFF
--- a/src/mame.c
+++ b/src/mame.c
@@ -617,8 +617,7 @@ static int vh_open(void)
 	/* if we're a vector game, override the screen width and height */
 	if (Machine->drv->video_attributes & VIDEO_TYPE_VECTOR)
     {
-       /*scale_vectorgames(options.vector_width, options.vector_height, &bmwidth, &bmheight);*/      
-       /*Hack to avoid segfault: leave vector resolution to its default the first time scale_vectorgames is caused*/
+       scale_vectorgames(options.vector_width, options.vector_height, &bmwidth, &bmheight);
     }
 	/* compute the visible area for raster games */
 	if (!(Machine->drv->video_attributes & VIDEO_TYPE_VECTOR))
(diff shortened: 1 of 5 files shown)


markwkidd:
I couldn't help at least trying these core options out a little more. Everything seems to hook up to the mame2003-plus code well, nothing is blowing up.

I've fine tuned the beam width core option values a little bit from the original patch but this would be an area worth revisiting in terms of what the best default values are for each of the vector options now.

markwkidd:
Thanks @KMFDManic ! This patch seems to be working well on a pretty wide variety of platforms. Woot! 🎉 🌈

KMFDManic:
@markwkidd I will most certainly prioritize a showcase of this addition for my Next Release.  Last Update,  I spread word that MAME 2003 Plus Experiemental was one of the most actively worked on Arcade Cores, as well as detailed the 100, give or take games that are new to coding, versus .78:)

nayslayer:
@markwkidd Skimmed the code, everything seems to be in order. One other problem with high resolutions is that MAME's menus (the tab-menu, which for some reason works only on the non-plus MAME) and messages aren't scaled. This is something I'll definitely look into. This one turned into an extra nasty nest of bugs, huh?

markwkidd:
@nayslayer to access the MAME menu in mame2003-plus you can either activate it as a core option, or you can turn on either the `mame_keyboard` or `simultaneous` input mode and press `Tab`. That is one of the differences between the two cores.

Thanks for sticking with this project a little longer. There do seem to be a couple of new quirks that have sprung up but IMO nothing at all that calls into doubt  your implementation approach. :)

PR #411, opened 2018-10-20
indicate core restart for antialias option, sync default vector options

markwkidd:
This clarifies that the core itself must be restarted to enable vector antialiasing to avoid confusion with resetting the content.

This also syncs up the descriptions and default values for the vector options between mame2003 and mame2003-plus where they had diverged a little bit. I'll be putting in a PR in for this purpose in the other repo too.


markwkidd:
I snuck in a tiny bit of core options whitespace tweaks in this commit as well apparently :)

PR #416, opened 2018-10-23
Fix double joystick inputs

grant2258:
we can continue discussion in the thread


grant2258:
@markwkidd is reaady to pull defaults can be left to people that use this setup

grant2258:
I dont know what going in with git hub atm mark but commit 

 make enough room for 6 players

didnt make it on the changes

PR #418, opened 2018-10-24
inialise retrojsstate to zero since we only read 4 joysticks data

grant2258:
the part of the pull request that I somehow messed up


grant2258:
no problems at all mr mark

PR #419, opened 2018-10-25
fix off road for mr widd

grant2258:
hi mark this ones for you it an issue you wanted fixed.

First we stop analog controls taking over all the time.

I mapped it to mame button 6 r2 or r on gampads depending on which controller type you decide to take over.

if you want to use the analog accelerate its rstick up but the digital button are fixed too. 

Enjoy


grant2258:
I just want to add this is a significant update all games that used analog controls previously did not allow digital controls to work because of the way it was added before

markwkidd:
Sweet! First, the implications of this are large! But also, Off Road was always what I played at the movie theater growing up 🥇

Wilstorm:
Ah, ok, yes at the hardware upgrade screen between races I had a hard time selecting because it scrolls so fast you overshoot until you finally get it after a few tries. Can you adjust dial settings for a controller or dial/mouse only?

grant2258:
joy/key only effect digital (keypad/keys the mouse is analog so all you can change is the sensitivity

Wilstorm:
Ok, that sounds good. I think this will work with Stun Runner too. I am not sure what other golden classics this will work on but it seems it will open up a whole new wing in the library! ;)

PR #422, opened 2018-10-26
add the buttons to stick hack back

grant2258:
ill add this properly when we add the analog controls in but we should leave it functional

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -1511,7 +1511,7 @@ int get_mame_ctrl_id(int display_idx, int retro_ID)
   return 0;  
 }
 
-#define DIRECTIONAL_COUNT           4  /* map Left, Right Up, Down as well as B, Y, A, X */
+#define DIRECTIONAL_COUNT           8  /* map Left, Right Up, Down as well as B, Y, A, X */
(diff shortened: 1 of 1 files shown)


Wilstorm:
Ha, so, well I already went through and remapped all the incorrect inputs and everything works fine, time to undo it. But after this update I should delete my ROM configs for those games and the defaults will work properly? Thanks I guess I had no idea it was a quick fix and have accepted weird inputs as is for some games.

Wilstorm:
I'm not sure if the update was meant to address Super Smash Bros., Total Carnage and Cloak & Dagger right stick firing but it's still the same issue but just reintroduced double inputs in the MAME TAB. I can open a new issue as out of the box these games are not playable using default controllers mappings per the Retroarch diagram.

grant2258:
Am at work will look when I get back

Wilstorm:
Thanks good sir and have a good weekend!

grant2258:
@Wilstorm some changes we do require a clear of default mame mapping configs it saves. It doesnt happen very often this is one case it does need done. 

The pull request is in for the fix it was a simple fix just had to get to work so pushed it without testing.

PR #426, opened 2018-10-27
fix the rstick hack causing double inputs

grant2258:
tested working any problems let me know


grant2258:
done pull this yest making some changes will probably just map this in properly and remove the hack

PR #428, opened 2018-10-27
map rstick properly

grant2258:
@markwkidd  we are all done I havent removed the rstick to buttons option  because this is mapped in now. 

It defaults to the rstick for these defaults if you still want the buttons to work we will have to hack them in not hard to do if that option is still wanted and rename the hack to rstick to buttons and add the code needed.


grant2258:
rstick

Issue #301, opened 2018-07-03
Rootbeer Tapper and Tapper High Scores:


Loggahead:
Is there an issue with highscores not saving on rbtapper and tapper?  Not sure if this is something new or something you already knew about but I can confirm on my rpi3b running retropie 4.4 I found that neither of those roms are saving highscores.

grant2258:
not all games save hi scores if no games are saving high scores that would be an issue

grant2258:
@markwkidd fwiw I believe the changes to the mame re-mapper caused this issue because the .cfg files arent saving either even when it enabled.

grant2258:
actually scrub them comments teh hi scores and cfg is saving forgot retropie uses the content dir duh

Loggahead:
Also for what it is worth, it appears to be affecting just Bally Midway games.  I confirmed that Timber is also not working.  I haven't tried some of the others like Rampage, Spy Hunter, etc. yet.

grant2258:
I think all the mcr3 will be effected at least mcr1 should be fine though

Loggahead:
In your opinion is it a fixable issue for mame2003+?  I'll do some testing and see if it also happens in mame2010 to see if it is just happening in 2003+.

grant2258:
all i can do is test with later on when i get round to it ill know for sure when i change the code. Is fba saving midway hiscores ?

Loggahead:
Not sure about fba.  I'll test them out tonight with in .43 if they are available and will post back.

markwkidd:
@Loggahead can you confirm whether this is still a problem in the latest builds?

grant2258:
nothings changed in respect this so its still and issue no ones fixed it

markwkidd:
I can confirm that scores are still not saving in `tapper` in the current build. Tapper did save its own score tables, so I believe these scores should be saved in the `tapper` `nv` file that is generated.

mahoneyt944:
Fixed https://github.com/libretro/mame2003-plus-libretro/pull/780

mahoneyt944:
This issue can be closed now as it's now fixed.

QPCloudy:
> This issue can be closed now as it's now fixed.

Hello. running Mame2003 with 0.78 rom set, Tapper and TapperA both are not saving the high score list. It does not appear to be fixed. Any suggestions?

mahoneyt944:
[...]

PR #432, opened 2018-10-27
DO NOT MERGE ignore only control mappings in the cfg

markwkidd:
My current implementation of the "Activate MAME Remapping" option is very simple -- when active, this option tells the MAME mode that there is no CFG file regardless of whether that is true.

However in addition to control mappings, the MAME CFG files also include information regarding coin and ticket counters as well as metadata regarding the internal audio mixer. I'm not honestly sure that this core uses any of that information but it seems better to be a little more sophisticated in my approach.

@grant2258 I feel I must turn to you as someone who is experienced with CFG files. Does this change make sense to you?

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -176,6 +176,10 @@ static int input_port_read_ver_8(mame_file *f, struct InputPort *in)
 {
 	UINT32 i;
 	UINT16 w;
+  
+  if(!options.mame_remapping)
+    return 0;
+  
 	if (readint(f,&i) != 0)
 		return -1;
 	in->type = i;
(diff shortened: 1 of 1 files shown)


markwkidd:
One reason I got interested in a making this core option more sophisticated is concerns like in the Tapper hiscore issue: https://github.com/libretro/mame2003-plus-libretro/issues/301

Could it be that the Coin Counter data or some part of the CFG file other than the remapping data can affect high score records? https://github.com/libretro/mame2003-plus-libretro/issues/301#issuecomment-402371581

grant2258:
its not that that is causing the issue I looked at future code where it works. I forgot all about it will look again when i have time I just forgot all about it. I will enable the mapping just in case.

markwkidd:
closing this as misguided on my part 🌞

PR #438, opened 2018-10-30
fix rstick option so it works remove input interface as its not needed

grant2258:


grant2258:
this rstick option was re implemented for people with snes pads incase you are wondering why its back. Its in its propper place now though. So you can use the rstick or buttons

markwkidd:
Pardon me I have edited your title - I see that you are still planning to do a test so I didn't want one of us to accidentally merge the PR prematurely

grant2258:
No the test is something different is ok to pull this. It will have no effect on the test

grant2258:
I'm pretty sure the the other option just delete the config and puts a map in

markwkidd:
got it, thanks and merged

PR #442, opened 2018-10-31
fix pacmania buffer overflow

grant2258:


grant2258:
https://github.com/libretro/mame2003-libretro/issues/381

PR #450, opened 2018-11-03
the ipac solution attempt

grant2258:
Im going to Keep this simple as possible, because it really is.

The config menu has been changed L3 on the controller to set this to the tab (or any other key you like) key in  RA->Settings->input->input user1 binds and set this to whatever key you want to bring the tab up with.

Next step is to stop the RA hotkeys causing issues there are two ways to  do this nul them or the more sensible way set hot key enable (i use numlock) this basically means you have to press numlock and the hotkey to get you menu up.

you set this in RA->Settings->input->hot key binds->enable hotkeys

now at this point your ready to go setup you user1 one binds  user1 and user2 binds in RA for you keyboard or ipac. 

congratulations you ipacs treated like an retropad like it should be from the ra config.

want to use the keyboard in full access mode no problems. goto options set input to mame go back press the game focus key (all keyboard binds go even retropad)   and  your done


grant2258:
@arcadez @markwkidd can one of you pull this we can get some testers on the retropie forum no doubt

grant2258:
tah

PR #456, opened 2018-11-04
Added A7A7 makefile target

markwkidd:
I'm going to go ahead and add this to the mame2003-plus Makefile -- knowing that there is a slowdown compared to mame2003 I see only upsides to at least having a functional recipe here even if it does not result in a core that runs at full speed.


grant2258:
what in particular is slow is it just everything?

PR #457, opened 2018-11-05
add marks keyboard only back

grant2258:
there does seem to be a case use for this as well mark so ive added it back


markwkidd:
Because I've been trying to follow discussions, I know that this is completely doable in RetroArch without the core option but also I think it's OK to have both available.

I'd eventually like to help make another small set of docs that aren't purely about the core like the official libretro doc, but are about using the core within RetroArch specifically. Basically a user guide focusing on mame2003-plus with RetroArch. That would be a place to suggest "best practices" like using the RA focus mode, and remapping controls from the frontend instead of the MAME menu.... todo...

grant2258:
yea im trying to steer away form that direction in the forums

PR #459, opened 2018-11-05
NOT READY YET improve legacy remapper core option

markwkidd:
This new approach to the MAME/Legacy control remapper core option is almost ready. @grant2258 this is my attempt to implement the concept I was describing over the weekend.

The purpose is to allow the CFG files to be saved so that dipswitch settings are maintained, but to do so in a way that still helps new users avoid making changes via the MAME remapper that they don't want or understand.

Because this new approach does make it, in my estimation, a little easier than before to accidentally create an unwanted CFG, I believe the answer is to make it as easy as possible for the user to get back to a clean slate. With this PR, the core option only controls what submenus are accessible from the MAME UI. Turning on the `Legacy Remapper` will give access to the mapping submenus as well as two new entries I've created:

* Flush current CFG
* Flush all CFGs

### Remaining issue
At the moment, things are close to working. However when the current game's CFG is deleted, the current modified mappings are still in active memory. Then when the core exits, these modified values are written back to disk, defeating the purpose of deleting the CFG file.

I've already worked out how to avoid this problem for the `default.cfg` file so I probably just need fresh eyes to get things completely working.

I decided to go ahead and open this PR because even as things are now it should be possible for grant or anyone else who is interested to load this version of the code, see the UI, and hopefully understand how this is supposed to work.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -288,10 +288,10 @@ static unsigned int count_input_ports(const struct InputPort *in)
 
 config_file *config_open(const char *name)
 {
-	if(options.mame_remapping)
+/*	if(options.mame_remapping)*/ 
     return config_init(name, 0);
 
-  return NULL;
+/*  return NULL;*/
 }
 
 
@@ -302,10 +302,10 @@ config_file *config_open(const char *name)
 
 config_file *config_create(const char *name)
 {
-	if(options.mame_remapping)
+/*	if(options.mame_remapping) */
     return config_init(name, 1);
 
-  return NULL;
+/*  return NULL;*/
 }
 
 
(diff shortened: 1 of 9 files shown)


markwkidd:
This is how it looks:

screenshot 2018-11-05 13 09 49

markwkidd:
Let me think about how to make the two 'CFG delete' commands more intuitive.

Maybe people won't turn on the Legacy Remapper at all so they won't need the delete commands and we can wait to see if that really would be useful for troubleshooting.

Because everything else is apparently working I will merge un update to this PR that makes the other changes from the branch but doesn't enable the delete functions. That way dipswitch changes will save for everyone, at least.

PR #461, opened 2018-11-05
fix metal slug possible other games unknow as of yet

grant2258:


markwkidd:
makes sense, thank you

PR #462, opened 2018-11-06
no functional change - optional for core option to create mame2003 subfolders

markwkidd:
One way in which MAME 2003 had to be ahead of the curve for most other libretro cores is in creating a primary subfolder in the SYSTEM and SAVE folders where all of its files can tidily exist separate from the SYSTEM and SAVE content from other cores that are installed alongside it.

Now in 2018, it is more clear that this is a scenario -- lots of core-generated files, some with filenames that can conflict with the filenames used by other installed cores -- needs a more standard solution and in fact there is one. RetroArch has already added an option to sort save files within a subfolder named after the core.

Without this PR, if a user has RetroArch configured to sort their save files, the result is a nested pair of mame2003-plus folders... not so nice.
screenshot 2018-11-05 18 03 21

I think one day there will also be an option within the RetroArch frontend (and other frontends) to sort their SYSTEM files this way too so I've pre-emptively added an option to disable that within the core as well.

This PR allows people who are already using the RetroArch option to sort their saves within named subfolders to have a sensible experience with this core. It also allows those who have installations built around the "legacy" behavior to continue on as-is with no change. Eventually I'd be interested in changing these to default to `disabled` but I don't feel there's a rush for that.

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -181,6 +181,21 @@ int osd_get_path_count(int pathtype)
  *****************************************************************************/
 void osd_get_path(int pathtype, char* path)
 {
+  char save_path_buffer[PATH_MAX_LENGTH];
+  char sys_path_buffer[PATH_MAX_LENGTH];
+
+  save_path_buffer[0] = '\0';
+  if(options.save_subfolder)
+    snprintf(save_path_buffer, PATH_MAX_LENGTH, "%s%s%s", options.libretro_save_path, path_default_slash(), APPNAME);
+  else
+    snprintf(save_path_buffer, PATH_MAX_LENGTH, "%s", options.libretro_save_path);
+
+  sys_path_buffer[0] = '\0';
+  if(options.system_subfolder)
+    snprintf(sys_path_buffer, PATH_MAX_LENGTH, "%s%s%s", options.libretro_system_path, path_default_slash(), APPNAME);
+  else
+    snprintf(sys_path_buffer, PATH_MAX_LENGTH, "%s", options.libretro_system_path);
+    
    switch (pathtype)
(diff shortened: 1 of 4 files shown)


grant2258:
ive never had any issues you pointed out above how do i recreate mark? RA will save on contents if you have no save folder set. I think the directory problems you seeing its your online game commits causing it

markwkidd:
I think you would have never had a reason to turn on this setting in
retroarch, plus it's relatively new.

Follow the steps in this post if you want to try it out:
https://forums.libretro.com/t/saves-in-difrent-folder-option-solved/9409/4

On Nov 5, 2018 6:53 PM, "grant2258" <notifications@github.com> wrote:

ive never had any issues you pointed out above how do i recreate mark? RA
will save on contents if you have no save folder set. I think the directory
problems you seeing its your online game commits causing it

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/libretro/mame2003-plus-libretro/pull/462#issuecomment-436113824>,
or mute the thread

grant2258:
makes sense now  mark if someone has it on cheers for that.  The problem is will APPNAME ever change again ?

markwkidd:
I've been wondering if we're ever going to get that letter from MAMEdev
about a rename....

grant2258:
I think ill check that this code out.I take it that it adds the core folder name to the savepath instead if the save path itself

markwkidd:
Yes I think you have the gist of the RetroArch feature -- the frontend modifies the libretro save path to include a subfolder named after a core, so that is already part of the save path that mame2003-plus gets from RetroArch.

PR #463, opened 2018-11-06
fix scrolling in history so it works and fix display game info

grant2258:
@markwkidd this broke the displaygame 2188c85effda827a72e9898e47843fb041009043 I have no idea what the intention of this code was.  

The history text can now be scrolled not sure where it went bad in the commits but is fixed now. 

The code is hacked in so many places it not even funny anymore lol


markwkidd:
There was a reason for this but if this PR works on your end then I'm glad to assume the underlying issue is gone now. Ill to ahead and merge

Issue #464, opened 2018-11-07
cheat-related controls issue


grant2258:
There was a post on the froums about these games. I did ask for peoples setup which i never got but looked it up myself in the end.

What is happening this is a dual joystick games but a hack has been added for single joystick control. 

The problem we have is mame Lstick and the mame p1 joystick both point to the dpad and the hack will mess with dual stick control although the game seemed to up down left right fine just using the dpad itself.

possible solutions

1) add a option to set input to none in the tab menu

2)  map the lstick in and use the analog l/r stick for dual joystick games

3) remove the single player hack

2 isint a real viable option if you where using and arcade panel you would have the same problem

1 setting the key to none is an option on the table but not a real long term solution for the average RA user that doesnt want to read  the book of setting up mame controls

to me removing the hack and leaving the driver like it should be is the right solution here. 

Even though i think this is the solution for this game.  The mapping the lstick for digital use and the adding the set to none aren't off the menu thats something that can be decided separately form this issue

grant2258:
ok ive been digging in a big more and found on the original mame78 the only way the p1 up/down/left/right activates to any binding is when you start with mame with the -cheat on the command line. It seems to always be activated on out port. So it would only have this problem if you had cheats on. 

https://github.com/libretro/mame2003-plus-libretro/blob/240ef74756167e9d6bfc5821b10391f05f5eaded/src/drivers/namcos2.c#L1372-L1378

now here is the effect of this in our dipswitches
https://pastebin.com/hdk7vrMx

now here is the controls that could be effected 
https://pastebin.com/TXmR15Zm

My suggested remedy it return the cheat on off as an option in the core options and remove any hacks if any are related to this change.

grant2258:
commit that changed the input ports behavior https://github.com/libretro/mame2003-plus-libretro/commit/fabdfac34c219ae9ccea191ab4d45d30c2f9d69c#diff-c7698bf8bb6ee8a6035a02a296c7a9b4

grant2258:
i would help if i pointed you in the diff direction sorry about that

 https://github.com/grant2258/mame2003-plus-libretro/commit/439ac137c7aa7f6429b409c124fa5dd2266429f1

grant2258:
@markwkidd the pull request is in to fix this. 

All the relevant information is here a condition on the cheat ports input  ports wasn't checked was all. This could effect many games and its dealt with at the level that fixes the problem. It has the potential to mess with many games. 

[...]

Issue #451, opened 2018-11-04
todo mame ui clear key_behavior choice


grant2258:
see topic here .

to me what we have is the correct behavior ie it sets it do default. 

When you select a new key it changes and to get your default back you just do the magic key press.

Some people want it to set to none which makes the point of having defaults useless to me its a great feature you can change it or get the default back. If you want it TO set  none you should be setting your general default to none.

grant2258:
just for reference i tried this in mame078 the behavior is identical

grant2258:
@markwkidd the code change to set input(this game) to none are here I still can see a real use for it but the team can decide what they are going to do. 

https://github.com/grant2258/mame2003-plus-libretro/commit/5026078816fcc6e2f1ee6a68ca710f8f937d3e2a

markwkidd:
As you know I've posted in the retropie forum thread, although I should probably post again there. I'm not all that excited about this change.

It seems like with this change you basically "lose" access to the default once you customize a game, and once the default is gone you have no way to get back to it from within the core unless you remember what it was or research the answer.

One way that I could imagine the MAME remapper UI being better would be if it had a command at the top of the mapping submenu that resets the mappings back to default. I might be interested in adding a `Reset All to Defaults` action at the top of the menu like that as an alternative to my `Delete current CFG` concept.

grant2258:
it can be a core option or you can put a toggle in the code im still waiting on solid information on this

markwkidd:
Somehow that conversation got off on the wrong foot, but I too would like to know more.

Right now I can't see a situation where the `None` mapping would be needed except as a workaround to a bug but I am willing to believe. I need to reread the thread but I put my energy for now into Dip Switches I think.

grant2258:
i believe mark will be taking care of this when he get the time. ill close for now since its more a todo thing now

PR #465, opened 2018-11-07
disable cheat inputs by default until a core decission is made

grant2258:
@markwkidd this is the fix simple and effective. The other issues of the this topic might have some merit the reason for the bad input is fixed. I think a simple core option to turn on or off will be suffice but it should be off by default with undefined behavior.

### DIFF
--- a/src/inptport.c
+++ b/src/inptport.c
@@ -1826,6 +1826,7 @@ InputSeq* input_port_type_seq(int type)
 
 InputSeq* input_port_seq(const struct InputPort *in)
 {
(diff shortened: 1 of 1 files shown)


grant2258:
https://github.com/libretro/mame2003-plus-libretro/issues/464
https://github.com/libretro/mame2003-plus-libretro/issues/451

grant2258:
assault

grant2258:
just to be clear mark the cheats still work this is just input side of things like dip switch settings ect and activating hacked controls

PR #471, opened 2018-11-08
Improve control detection, particularly 4-way, 2-way, and rotated joysticks

markwkidd:
As often the case, this turned out to be a little more coding than I expected. Some refactoring was necessary, and then some more refactoring also seemed prudent.

This whole PR should have no effect for the user but since it touches a lot of lines I wanted to at least tag @grant2258 since this touches some of your recent code about the 4-way joystick core option.

This PR:
* **DOES** add detection for 2-way and 4-way joysticks
* **DOES** only show the 4-way core option if users have either a 2-way or 4-way joystick
* **DOES** add a new `bool` called `rotate_joy_45` to the controls.c data structure which is used to flag games with rotate joysticks like qbert
* **DOES** set `rotate_joy_45` as `true` for qbert and its kin, as well as the `inferno` romset. Those are the only ones known by the controls.dat project to have rotated controls

This PR:
* **DOES NOT** automatically use 45-degree mode when that flag is true on your 4-way joystick core option @grant2258 because I wanted to check with you first


Wilstorm:
@markwkidd - Wow Mark, so even 2-way will be restricted? That's really fancy! One quick question. do we leave it on "original" or "new" and it will automatically use it for all 4-way, 45 rotated and 2-way? Does the 45 rotate also include Congo Bongo (congo.zip)?

markwkidd:
At the moment there is no automation for the 45-angle rotation in this code although I think that should be possible.

Thank you very much for pointing out Congo Bongo. I will need to add that 45-degree metadata to controls.c . Any other games you can think of that have rotated joysticks would be very helpful

markwkidd:
Alrighty, `congo` is now flagged as having a rotated joystick, thanks again Wilstorm

Wilstorm:
@markwkidd - Here's a list of what I could find in the Plus DAT. I added Jack Rabbit even though it's an 8-way joystick but only 4-way is used during game play. It probably doesn't count but I added anyway. Tip Top is a clone of Congo Bongo. The name is from the DAT and also the ROM is listed after.

```
Congo Bongo congo
Tip Top tiptop
Inferno inferno
Jack Rabbit (set 1) jackrabt
Jack Rabbit (set 2) jackrab2
Jack Rabbit (special) jackrabs
Q-bert (US) qbert
Q-bert (Japan) qbertjp
Q-bert (early test version) qberttst
Mello Yello Q-bert myqbert
Faster, Harder, More Challenging Q-bert (prototype) sqbert
Q-bert's Qubes qbertqub
```

markwkidd:
Thanks for `tiptop`, I've pushed that change.

>Jack Rabbit looks like the other diagonal games, but it definitely uses an 8-way stick.
Moving during game play uses only diagonals, but entering high-score initials uses **only** up/down/left/right.

According to this thread: http://forum.arcadecontrols.com/index.php?topic=151056.0


I've been looking around arcadecontrols.com and I'm seeing lists of "diagonal 4-way" and "rotated 4-way" games that make me feel like we have it all flagged now. For example in 2013 here is the list that ROMLister produced of "diagonal 4-way" games: http://forum.arcadecontrols.com/index.php/topic,129696.msg1326815.html#msg1326815

Wilstorm:
@markwkidd - Ok, I knew it played with diagonals but the high score needs the 8-way so it would need some special processing. I don't know if I ever even played it.

The list from Arcade Controls almost match. I actually pulled the games directly from the DAT generated by plus so it should be accurate. I know there's more games in later versions but I just focused what was in this version.

That's pretty neat. I guess I never thought about it but you have basically two types. Actual rotated joysticks to work at 45 and ones that are probably mounted straight but play on the diagonals.

markwkidd:
Now that I'm feeling more sure we have a good list of rotated joysticks I couldn't help myself, I've tried to implement automatic switching between regular 4-way and rotated 4-way. That code is here in the branch now but needs to be tested for sure.

markwkidd:
I'll add that I'm not positive that this will apply the 4-way code to 2-way games. To be continued...

grant2258:
I really dont mind whatever way you want to implement this mark. I just added this for robert and Wilstorm. Ironically i received my circular restriction plates from china.

markwkidd:
@grant2258 I sometimes try to add one more component to a phase of work than is needed. In this case it would be incorporating 2-way joysticks into the brew so I have removed that part.

grant2258:
ill try explain the issue here to you and why its not a good idea to change it. Mame by default in an input define will assign IPF_PLAYER1 if nothing is specified  because its define as 0 as well so is IPF_8WAY and IPF_4WAY

now lets look at the 2 different ways you can define and input for player1

PORT_BIT( 0x10, IP_ACTIVE_LOW, IPT_JOYSTICK_RIGHT | IPF_2WAY )

and 
PORT_BIT( 0x04, IP_ACTIVE_HIGH, IPT_JOYSTICK_LEFT  | IPF_2WAY | IPF_PLAYER1 )

if you change this define value you would need to also add the code to all the input routines to account for this. 

PORT_BIT( 0x10, IP_ACTIVE_LOW, IPT_JOYSTICK_RIGHT | IPF_2WAY )

you would also need to add code to the input functions to account for it

the 4way code needs distinguished because it needs processed differently in the input functions.

for information purposes up and down wont be set to differentiated it if you want to detect it

markwkidd:
I'm interested in whether it would be useful to adapt your code for 2-way use but it may be "a problem looking for a solution". Your current 4-way code is a big advancement for the core @grant2258 so I'm excited about making it as easy to use as possible.

I've refactored the PR to remove the 2-way joystick stuff.

If I'm just using a keyboard, is there a way for me to test that your 4-way restriction code is working properly?

grant2258:
yes there is a way to test it mark start dkong. walk right and press up he will keep walking left in the new code

markwkidd:
Well dangit... `qbert` is correctly identified as 4-way and as rotated joystick but `dkong` is still being detected as 8-way. Something is evidently incomplete where I detect the 4-way input mask

grant2258:
I have to go to work but i should be off the next 2 days hopefully will take a peek when i get time if you dont beat me to it

markwkidd:
Well, I took a short nap, had lunch, and found the silly logic error immediately. Now with `4-way` enabled, Mario keeps walking even if I also press the Up button.

Based on the way I can no longer control `qbert` at all with my keyboard on my 4 directional keys with `4-way` enabled, I feel pretty certain that the automatic rotate 45 logic is working OK.

I've pushed this fix. I'm open to doing other tests as suggested by the peanut gallery, or getting someone else to test this out as well. No need necessarily to merge it yet. It sounds like grant may have some time coming up soon.

Wilstorm:
@grant2258 - Thanks for the list. I did a cut-n-paste for reference for 4-way games.

grant2258:
i can pretty the list up for you and include full names as well if you like just let me know its no bother

grant2258:
@markwkidd you can test qbert on the keyboard as well up/left  up/right    down/left  down/right key  press them at the same time it requires a diagonal for input

markwkidd:
Alrighty, `qbert` is properly rotating automatically with the new core option.

It passes basic tests that we have for the 4-way code. I feel ready to merge this pull request. It's always a little nerve wracking to hit 'Merge' on a big PR -- especially when any bugs are going to be my responsibility -- but here we go!

PR #487, opened 2018-11-12
system16 update

grant2258:


markwkidd:
does this PR mean you've found a way to resubmit the patch? I'm hoping that's so

grant2258:
yea i just started over again luckily  i saved most of the changes there is no commit updated but they are all there/

ost for moonwalker and outrun. 
afterburner palette fix

markwkidd:
Are there any specific games that you would still like to have tested?

I saw you mentioned in the previous PR that you had figured out the After Burner palette issue -- what else can folks check out?

grant2258:
well to be honest I havent had time to test every game I cant see there being any issues to be honest. I guess if there is no testers just pull it. im sure we will hear back

markwkidd:
Yep, it's hard to guess how many users this core has, but I do get the sense that the number is slowly but steadily increasing. It's not thousands and thousands, but here, on discord, libretro forums, and retropie forums I detect increased awareness of what's going on with this core.

At any rate, with a popular group of games like system16 I agree we'll probably get bug reports if there are regressions sneaking around. Thank you!

grant2258:
it was all @arcadez  work i just brought it back in

Wilstorm:
If you have a list of games I could do some testing.

grant2258:
well it would need to vs the old driver as well. But you can test the old driver in mame2003 as well. Ill try make a list up when i get time. Well ones that human readable without the needless info

Wilstorm:
Ok that sound good.

arcadez:
Well it was busted before ya fixed firstly the inputs and then a pallette issue in After Burner, so
i guess we can both take some credit for this rather large update :)

arcadez:
A quick breakdown of some of the improvements via this update would be.........

prelim vdp graphical emulation which fixes tile banking for Shadow Dancer
bad sprites and tiles fixed for Heavyweight Champ and Wrestle War
all versions of Tough Turf now have full sound although sprite issues remain

grant2258:
Ive looked up this driver it seems to me our best option is to used the unprotected sets for the main cpu 68000 or z80 sound depending on what protection it uses. This is how they fix broken arcades the just replace the roms needed and put a normal 68k/z80 in then fix whatever gfx sound issue we can

PR #488, opened 2018-11-14
add core option for cheat dipswitch/input port enable/disable

grant2258:
Im not sure if you want this as a core option or just off if its just off ill just remove the core option part of the code just let me know. ill add screen shots of the differences with the assault thisgame menu. 

Ive also added some parts of the input code that needed fixed up where it was removed before when the cheat flag is active,

I also added the code for the dipswitches as well to show/hide the cheat input code back and added the code that was missing from input this game to do the same as the dip-switches this was missing in mame078.

assault-181114-064024
assault-181114-085047


markwkidd:
I'm going to go ahead and merge this -- it's definitely an improvement over the current state. I plan to educate myself on how useful these added input actually are in real life but for now it seems like a no brainer to at least be able to access the functionality

PR #498, opened 2018-11-27
try fix Fileio for android

grant2258:
Hopefully this does the trick thanks to @radius


grant2258:
just updated the core on retroarch ist at 88ab4def62c7e8bd9b35e493e44eb4ea66d422f7 ads seems to be working again before this pull request. Very starange

grant2258:
this seems very temperamental it creating dirs as mame2003- and sometimes mame2003-plus. I guess ill need to look into this more somethings not right here the android end. Ill need to figure this android stuff out

PR #506, opened 2018-12-02
Remove nrallyv hack

markwkidd:
This is another product of my metadata project. In MAME 0.78u1 `nrallyv` was removed as a simple graphic hack.

> 0.78u1: Stefan Jokisch added proms ($120, 140 - vertical sync). Removed New Rally-X (vertical hack), because the graphic hack (rotated tiles + display) not worth supporting.

I'd propose to do the same in this core by commenting out the `nrallyv`. Maybe if there is an `#ifdef HACKS_MODE` one day we could put this driver in there. Is this OK with you all?


grant2258:
no a problem removing it if you want mark

PR #511, opened 2018-12-04
add ssonicbr driver

markwkidd:
from @arcadez and @grant2258 

I'll work on the merge issue since it's due to my metadata updates.


grant2258:
i can fix this my end if you want

Issue #440, opened 2018-10-30
use the frontend to display 'missing samples warning' 


markwkidd:
Now that @grant2258 straightened out the logic that was segfaulting with incorrect romsets it got me 
thinking.

Sometime, maybe when I've got a three day weekend, I'd like to see if I can display a MAME UI/visual error message when the romset is wrong in addition to writing an error message to the log. Not a high priority improvement, but it would be cool 🕶 

edit: bparker06 from RetroArch has used the nuklear barebones UI library to add status messages to a different core. I don't remember the details although I think it was a PSX emulator core. I'll try to find out. This is nuklear: https://github.com/vurtun/nuklear

markwkidd:
hehe @grant2258 what do I see when I log into Discord this morning? You are talking about this same concept in #programming

markwkidd:
r5 in the #programming channel mentioned this additional information: 

> bparker used ugui on beetle-psx. I think it writes a message directly to the core's [framebuffer]
https://github.com/libretro/beetle-psx-libretro/tree/master/deps/ugui

grant2258:
well we might just need to adjust our view of the mame ui no point in using something else when mame has it built in.

grant2258:
I do still firmly believe ra should be capable of showing a dialog box with information and a yes/no action maybe im expecting too much but ill implement it like we discussed on #programming for now.

markwkidd:
I just had my mind blown by the news that the libretro API already allows us to display status messages via the frontend. the callback is `RETRO_ENVIRONMENT_SET_MESSAGE` which I'll definitely be looking into for use in this core!

grant2258:
https://github.com/libretro/mame2003-plus-libretro/blob/3701ec3497e8439c28e2d43d59d216ceff255a6e/src/common.c#L1344-L1363

this is where you can add the messages

grant2258:
there is one little request i do have can you leave the log_cb in as well when you update this.

markwkidd:
We should definitely keep log_cb intact.

Note to self: also use the OSD to display brief warning for romsets that are have a status of preliminary, baddump, nodump, or  protection.

grant2258:
Well this reminds me something does need fixed up. Ill tell you how to trigger this segfault make a new empty file call it boom.zip load it up youll get  setfault. 

Now rename this empty zip to wboy.zip load it it will say you have missing roms.  (wboy) is a valid driver name.

[...]

PR #524, opened 2018-12-09
fix segfault when gamename does not exist in mame

grant2258:
This was a particularly bad bug ive known about but seemed to always forget about never got round to fixing it. 

The problem is if the game name did not match out driver list we would segfault. This shouldn't happen ever.

sets change names if fba and different mame versions this will give us the failed to load contents message. Now this is fixed we know if we get a bug report of a crash its really a crash.

please see 
https://github.com/libretro/mame2003-plus-libretro/issues/440#issuecomment-445488380

about the boom.zip


markwkidd:
nice! thank you

grant2258:
no problem sorry i took so long just kept on forgetting about it. Ive know about this for a while.

PR #528, opened 2018-12-11
restore Nebulous Bee bootleg

markwkidd:
Nebulous Bee was removed in 0.79u1 as being a hack, but it is in fact a real bootleg PCB that is found in the wild.

It's been added back to MAME and so I thought I'd try adding it back here. I think this was a straightforward "re-add the ROMs and tweak the Nebulous Bee driver declaration so that it uses the new `galaga` init" ... right?

@arcadez and @grant2258 -- does this look about right? the game seems to run ... just want to make sure I've not just added code that will explode the core :)

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -512,7 +512,7 @@ const struct GameDriver *test_drivers[] =
 	DRIVER( galagao )	
 	DRIVER( gallag )	/* bootleg */
 	DRIVER( galagamk ) 	
-//	DRIVER( nebulbee )	/* hack */
+	DRIVER( nebulbee )	/* hack */
 	DRIVER( gatsbee )	/* (c) 1984 Uchida / hack */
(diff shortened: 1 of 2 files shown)


grant2258:
If the romloads match the new address maps should be fine. Look at the dkong fixes i done for donkey kdongx and skonpe you would need to test it t be honest mark

markwkidd:
It runs OK here but I'm not getting sound. It might be my setup though.

markwkidd:
Now that I have an audio driver that works better, I'm getting sound for Nebulous Bee and the Galaga fast shoot hack I added. Maybe things are OK as they are? Could I be that lucky? 🥇

grant2258:
just make sure you update the dat and run it through clr mame pro and fix and problems if they are any conflicts

PR #532, opened 2018-12-12
flag ichidntb as not working

markwkidd:
I don't think it is supposed to be working with the code we have now. Any thoughts that it should be left as is? If there is no objection I'll merge this in a couple of days.


markwkidd:
I've created a PR to flag it as not working: https://github.com/libretro/mame2003-plus-libretro/pull/532

PR #544, opened 2018-12-17
enable backdrops by default and re-add warrior artwork

markwkidd:


grant2258:
mark the artwork is in metadata as well you should delete any one of them you think is right to delete.

PR #554, opened 2018-12-19
Fixes and cleanups 

grant2258:
the dcs speedhack have been fixed for you mr @KMFDManic  when  this is pulled that is


KMFDManic:
Awesome:) Will test after work.

PR #555, opened 2018-12-20
add hard head 2[arcadez]

grant2258:
@arcadez is added my good man hope I done the change log ok for ja


grant2258:
@arcadez can you pull this when your ready bud

PR #559, opened 2018-12-26
Frogs backdrop

UDb23:
For 1080p (or 16:9 display); adapted from current MrDo's mame artwork.


UDb23:
Basically was not playable in mame2003 without the backdrop.
It would now be nice if someone could backport the sound support from more recent mame.

markwkidd:
Could you start an issue about the sound backport? That may get lost in
GitHub once I merge this PR

PR #561, opened 2018-12-29
fix false samples missing reports

grant2258:
flac will still most likely need fixed but wav is fine


markwkidd:
At the time I only checked for false positives not false negatives. Let me
see if I can merge this from my phone in a minute.

grant2258:
well I just checked flac is ok all is working like it should everything working as it should.

PR #570, opened 2019-01-05
standardize Makefile whitespace between platforms, a little tidy

markwkidd:


grant2258:
removing -std=gnu90 you going to cause problems

Makefile
```diff
@@ -545,17 +545,22 @@ else
...
 endif
 
-# Compiler Flags For All Platforms #############################
+# End of architecture-specific flags ######################
+
+# Compiler flags for all platforms #############################
+
+# we actually code to c89 but in order to allow // comments in c source, gnu90 is the closest standard
+# beware that gnu90 does still allow some code through that will not compile for msvc or emscripten though
+CFLAGS += -std=gnu90
```
markwkidd: @grant2258 gnu90 is relocated here

markwkidd: I guess that doesn't quite qualify as a tidy but. maybe "tiny refactor" 👾

grant2258:
will that work in msvc ?

markwkidd:
yes I think it is fine in msvc but actually not, as written, with the nintendo switch. let me think about how to fix my error

markwkidd:
I'm going to see if switch actually needs gnu11 or not.. time to watch the buildbot

PR #585, opened 2019-01-16
fix compiler warnings

grant2258:
There are far too many warnings in mame2003.c at the moment. That`s them all gone now

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -644,32 +644,28 @@ static void update_variables(bool first_time)
 
 void retro_get_system_av_info(struct retro_system_av_info *info)
 {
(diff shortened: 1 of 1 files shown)


grant2258:
@markwkidd some good news with the pointer fixes for the controllers it now saves the the game controller type on exit no need to save a remap any more

grant2258:
thought the same to be honest but looks like the last one i fixed actually fixes it. i want to start a clean slate for an attempt to fix the analog controls. I want you to keep an open mind when i do this.  From what ive looked at i can fix it to work but i have no idea how to deal with the ra fingers since some drivers change the default mappings in the drivers. There is no eta on this btw it might now work out but having working mapable analog would be of benifit

markwkidd:
I'll do my best to help. I did buy a used controller with an analog
joystick so I can do testing or whatnot.

PR #588, opened 2019-01-17
fix segfault when loggin is on

grant2258:
Made a little mistake by guessing what the code does. I found the issue when I put logging on and  then debugging the code. The pointer is fixed properly and the controls still save on exit sorry about that.


markwkidd:
it's all good

PR #590, opened 2019-01-17
add analog support

grant2258:
just in case somone want to test if there is enough interest to finish this off and fix the digital hats to work again as well.

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -1686,116 +1686,116 @@ int get_mame_ctrl_id(int display_idx, int retro_ID)
 #define PER_PLAYER_CTRL_COUNT_NO_DBL_NO_MOUSE   (DIRECTIONAL_COUNT_NO_DBL + BUTTON_COUNT_PER)
 
 #define EMIT_RETROPAD_CLASSIC(DISPLAY_IDX) \
-  {"J"   #DISPLAY_IDX " HAT Left",     ((DISPLAY_IDX - 1) * 26) + RETRO_DEVICE_ID_JOYPAD_LEFT +1000,   JOYCODE_OTHER}, \
-  {"J"   #DISPLAY_IDX " HAT Right",    ((DISPLAY_IDX - 1) * 26) + RETRO_DEVICE_ID_JOYPAD_RIGHT +1000,  JOYCODE_OTHER}, \
-  {"J"   #DISPLAY_IDX " HAT Up",       ((DISPLAY_IDX - 1) * 26) + RETRO_DEVICE_ID_JOYPAD_UP +1000,     JOYCODE_OTHER}, \
-  {"J"   #DISPLAY_IDX " HAT Down",     ((DISPLAY_IDX - 1) * 26) + RETRO_DEVICE_ID_JOYPAD_DOWN +1000,   JOYCODE_OTHER}, \
-  {"J"   #DISPLAY_IDX " B",            ((DISPLAY_IDX - 1) * 26) + RETRO_DEVICE_ID_JOYPAD_B +1000,      JOYCODE_##DISPLAY_IDX##_BUTTON1}, \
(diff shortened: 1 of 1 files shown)


grant2258:
Ok thats us fully functional you can map the dpad if you need to use it. One last thing I changed the Retropads to "R" originally but it could be mixed up with R1 R2 R3 so i changed it to J. 

Retropad is just too long for the tab menu. RP or J will suffice its whichever one is preferred.  I tested this on a few games havent came across any issues.

I havent found any issues yet on my 360 controller or xbox one. The pr is ready pending testing.

markwkidd:
If RetroPad is too long, please use `RP` as the abbreviation.

I'm deep in the MAME 2003 port process. I will test once I get through my next phase on that but you are moving pretty quick!

grant2258:
This is how it looks now I think it reads better with one letter.  I dont really feel strongly about it so you can put retropad back if you like. People can decide what way they want it at the end of the day.

offroad-190117-191120

grant2258:
ok changed look like this now. 

offroad-190117-193450

markwkidd:
looks great! would you like to have this merged?

grant2258:
I think we should have one or two tester goig by out bad luck at pushing things like this! I can provide instructions on how to compile on the pi

grant2258:
Ive been able to do this for a while what my issue is games like this.  LLander which maps the left right axis to up down and left right to digital left/right. 

Which is right for the game. And the RA fingers will show wrong for that game. Actually I dont know if analog is mapped at all in the fingers wont effect any digital games though. Just making you aware drivers can remap analog controls. Well it will now they actually work. 

hopfully @KMFDManic can test hard drivin now its map-able i have no idea were to gebing mapping that game i sure there was some guide mr k showed me

grant2258:
instructions to patch

for retropie
```
cd ~/RetroPie-Setup/
sudo ./retropie_packages.sh lr-mame2003-plus clean
sudo ./retropie_packages.sh lr-mame2003-plus sources
sudo patch -p1 -d tmp/build/lr-mame2003-plus < <(wget -qO- https://patch-diff.githubusercontent.com/raw/libretro/mame2003-plus-libretro/pull/590.diff)
rm 590.diff
sudo ./retropie_packages.sh lr-mame2003-plus build
sudo ./retropie_packages.sh lr-mame2003-plus build install
```

for *nix or msys then just compile as normal
```
git clone https://github.com/libretro/mame2003-plus-libretro.git
patch -p1 -d  mame2003-plus-libretro < <(wget -qO- https://patch-diff.githubusercontent.com/raw/libretro/mame2003-plus-libretro/pull/590.diff)
```

markwkidd:
@grant2258 I turned your build instructions into a slightly more generalized doc. Thank you for writing them up, not for the first time. https://github.com/libretro/mame2003-plus-libretro/wiki/How-to-compile-a-PR-for-testing

grant2258:
sounds like a plan its easier that way than setting github up right. You can pull this if its holding your updates back its easy enough to revert if you need too. Just make sure ive not undone any of your changes.

grant2258:
I wont be doing any changes unless its bug fixes i know its frustrating  with github at times

KMFDManic:
@grant2258 Excellent commit.  I knew you could pull it off.  This should benefit a lot of games, such as Cruis'n USA (I think), Hard Drivin', Race Drivin', Hard Drivin's Airborne.  I will test things out over the next couple of days.

@markwkidd This also makes me think of something else that might be interesting to "possibly" better optimize in some way, shape, or form.  If you run a Light Gun game, such as Vs. Duck Hunt, etc, with a controller, versus a mouse, sensitivity is very goofy and takes a little patience to get used to the nuances of how it is calibrated.  Some games tend to be overly sensitive.  I know there is generally a way to affect the overall sensitivity.  But, maybe, as a Core Option, mouse sensitivity with a percentage between 0%-100% or such can be directly changed from there, kind of like you can do with PC games such as Quake, Doom, and so on.  This would be a tremendous asset for those who don't necessarily have the luxury of running a mouse.

PR #593, opened 2019-01-18
 update discrete soundcore to mame080 

grant2258:
Ive updated the the sound core that should fix this issue hopefully. Our mame discrete drivers are very basic. 

Ive updated the 3 drivers that needed updating as well for the new functionality of the soundcore. The good news is we can update divers with discrete sound now since we have extra functionality in the core.

this is the games directly effected by this issue 
```
canyon
canyonp

atarifb
atarifb1
atarifb4
abaseb
abaseb2
soccer

skydiver
```


markwkidd:
There is a little build error that I think comes from this commit: http://p.0bl.net/82559

markwkidd:
Really nice PR by the way!

PR #603, opened 2019-01-19
add "Donky Kong Remix" hack

markwkidd:
courtesy of @arcadez , thank you very much!

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -791,6 +791,7 @@ const struct GameDriver *test_drivers[] =
 	DRIVER( dkongjo )	/* (c) 1981 Nintendo */
 	DRIVER( dkongjo1 )	/* (c) 1981 Nintendo */
 	DRIVER( dkongx )	/* (c) 2008 Bootleg  */
+	DRIVER( dkrdemo )	/* 2015 Bootleg  */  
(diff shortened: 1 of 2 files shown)


grant2258:
tis hasnt been been implemented right segfaults might want to change it to game not working until its fixed

arcadez:
It might be that the original dkong memory map is incorrect for this game.?? let's hang it off
machine init braze to test out my theory :)

arcadez:
Actually i just ported the exact same code across to MAME72 and the game booted and works
fine so the problem is on somewhere on this side, i'll revert that Braze machine init change now
as it's got nothing to do with dkrdemo not working.

grant2258:
i acttually forund teh fix was the gfx decode was fixed between 084 and 085

arcadez:
Hmm yeah good spot but it still doesn't explain why dkrdemo works for me in both the MAME72 and 84 xbox cores without the above graphical changes, it's a funny ole game this MAME malarkey.

grant2258:
i think my issue came with another heap overflow issue it was all the dkongs doing it. It never happened before I found both issues anyway. Only reason i picked this up is because i was debugging had two heap overflows

PR #606, opened 2019-01-20
Analog fix and audio skew bypass fix

grant2258:
as discussed earlier


grant2258:
dont pull this yet need to test mark

grant2258:
there is an issue with out run analog working with this on the dpad ill change thing about a bit. I think I know what it is.

grant2258:
ok mark this is good to go

Issue #400, opened 2018-10-15
[Bounty: $50] implement libretro lightgun API 


markwkidd:
## Current lightgun support

**MAME 2003-Plus 'supports lightguns' in the following way:**

1. The lightguns must be set in hardware or the operating system to emulate a mouse
2. We rely on undocumented RetroArch behavior that passes mouse input directly to the core even though this core does not implement the mouse input interface
3. MAME 2003-Plus has a lightgun input interface -- however we do not use this MAME code within the core, instead using the mouse input fallback codepath for lightguns

## Resolution

Connect the libretro lightgun API to the MAME 0.78 lightgun OSD API. The overall approach is to implement these features on both sides and normalize input values for MAME use.

MAME:
- osd_lightgun_read

libretro:
- RETRO_DEVICE_LIGHTGUN

markwkidd:
Even though I eventually want to add some bounty money to this issue myself and promote it to people who would be interested, I'm currently focused on hyping the RetroArch integral artwork bounty so this will have to wait its turn.

markwkidd:
Bounty up to $25 -- thanks @TillmanZ

grant2258:
I would do this if i had a lightgun!

markwkidd:
@nayslayer mentioned that they might take a shot at this

(pun intended) 😜

markwkidd:
Bounty up to $50 - thanks @Thatman84!

nayslayer:
Okay, I can probably do that - this seems easy enough - but is there anyone who can assist with testing? I personally don't own a lightgun (never even knew they can work on anything but old CRTs anyway), so anything I do will be (probably tedious) guesswork.

markwkidd:
I think you should be able to configure a mouse as a lightgun via retroarch
and the lightgun api. (Basically the opposite of how the core now
'supports' lightguns internally if they appear to be mice)

If you can't get a mouse to work via the libretro lightgun API in retroarch
then I'll work with you to figure out an alternative.

dougan78:
I am willing to test.  I have 2 light guns. (3 total mice inlcuding the tracbkall).
Mouse 0 Trackball
Mouse 1 Aimtrak 1
Mouse 2 Aimtrak 3

[...]

PR #608, opened 2019-01-20
WIP ignore only control mappings in the cfg

markwkidd:
@nayslayer pardon me for peeking into your repository, but I am doing a bunch of porting/synchronization between MAME 2003 and MAME 2003-Plus.

Specifically I'm now syncing some of the input code, and I was looking at your lightgun branch when I noticed this patch. Given that you didn't offer it to the core, this seems like an improvement over the current code.

Would you be OK having it added? If so, are there any issues with it other than the merge conflict from the new analog code?

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -176,6 +176,10 @@ static int input_port_read_ver_8(mame_file *f, struct InputPort *in)
 {
 	UINT32 i;
 	UINT16 w;
+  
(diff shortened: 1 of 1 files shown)


grant2258:
https://github.com/libretro/mame2003-plus-libretro/issues/400

grant2258:
this corrupts the files if you stop the right order

PR #623, opened 2019-01-23
add analog scale with deadzone

grant2258:
fr500 showed me some beetle analog code and i just added a negative check to it and scaled  it to +128 -128 values we needed so that should be us all finished up.

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -10,6 +10,8 @@
 #include <string/stdstring.h>
 #include <libretro.h>
 #include <file/file_path.h>
+#include <math.h>
+
 
 #include "mame.h"
 #include "driver.h"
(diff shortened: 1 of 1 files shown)


src/mame2003/mame2003.c
```diff
@@ -1893,11 +1895,34 @@ void osd_trak_read(int player, int *deltax, int *deltay)
 
 int convert_analog_scale(int input)
 {
-		int result;
-		result =  (input / (int) 326.68);
-		result = result * 1.28;
-		if ( ( result > 0 && result < deadzone) || ( result < 0 && result > -deadzone)) return 0;
-		return result;
+	int trigger_deadzone = (32678 /100) * 20; // 20% deadzone is plenty i would imagine
```
hhromic: Sorry to parachute on  your commit, should `32678` be `32768` (`2^16 / 2`) ? :)

grant2258: nope scales to 128

hhromic: Independently of the output scale, I thought the deadzone would be a percentage of the maximum input scale of `input`, which I assumed was `32768` as normally is in evdev and other input systems. Also later it's clamped to `32768 - 1` as well:
```C
		input = (int)round(scaled);
		if (input > +32767) 
		{
			input = +32767;
		}
			input = input / 326.78;
```
Hence my confusion because the number `32678` looks like a typo :)

hhromic: Just saw you edited your reply.. now reads "to a percentage".. which does make sense.. sorry for the nagging ! :)

grant2258: the rand is + / - 32768 not 16384

hhromic: But I still believe there is a typo because:
```C
input = input / 326.78;
```
will never be `1.0` if you are clamping to `32768`

hhromic: argh forget it.. Just realised my error in understanding... !  sorry again!

grant2258: its not going to make a difference the scale will still be 100 with 32768 or 32767

grant2258: can you please explain this 
```
Sorry to parachute on your commit, should 32678 be 32768 (2^16 / 2) ? :)
```

grant2258: you going somewhere else now

hhromic: You mean how I got that expression?
In evdev the analog range is `-32768` to `32767` because of using a signed short int to represent it. The range of a complement-2 signed short (16 bits) int is `- 2^16 / 2` to `2^16 / 2`.

Yes I realised I didn't read correctly the code and my point is invalid, so you can totally ignore all my comments to your commit :) Sorry again.

grant2258: no problems at i was going over to see if i missed something feedback is always welcome

PR #630, opened 2019-01-25
WIP Input system updates

grant2258:
Most of what changes is under the hood. The thing now is you can switch between legacy and none legacy without issues. If you do go to legacy it will set the default config to default as well as each game you play. The dipswich and coin counter information kept as well now when switching between modes. 

You must clear every file in mame2003-plus for this update. The structure is fixed now the internal mapping between pads wasnt great. Ive tested this and not had any issues.

analog enable on and legacy enabled will basically let you configure mame the way you would normally. 

analog disable is the default mode because some people only have arcade panels or  a pad with no analog.  

The rest works as usual. Any testing welcome it all does seem good atm with the test ive done mark but I want to test of the next two or 3 days when I get the time.


Wilstorm:
@grant2258 - Just to clarify. The defaults are analog enabled and legacy remapping disabled?

If we switch legacy remapping to enabled then the default.cfg file will be reset? 

So basically if we turn legacy remapping on it will be reset but if we shut it off to test something and turn it back on all our input changes will be wiped out?

grant2258:
no the defaults are analog disabled some people donr have analog sticks thats the reason.

If we switch legacy remapping enabled (ie mame default)  tab menu ect everything will work as it should and it will save all your changes like mame does.

If you go to legacy mapping off it will do this 

1) set your default.cfg keys to mame defaults.
2) every game you play with this on will reset to defaults when you exit. 

The none legacy mode does not have a tab menu so everything has to default for it to work no way round that

grant2258:
The only other solution is to save differnt config files for legacy and none legacy that would probably be more ideal. Ill wait for some feedback before doing anything new

Wilstorm:
Ok, I see what you're saying. Yeah I have to admit anytime you start wiping out configs it's doesn't sound good at all. 

Let's say someone is curious or troubleshooting or testing a digital controller or whatever reason they disable the core option, then all of a sudden they have to reconfigure all their defaults. As well as every game they changed from the defaults, it would be a nightmare. Some people have a lot of custom config setups, button values, etc. and one flick of the switch and it's all gone.

grant2258:
The don't have to re configure you can switch analog on and off it doesn't change your config file

grant2258:
The only thing that changes your config is legacy mode changing

Wilstorm:
>If you go to legacy mapping off it will do this

   >1. Set your default.cfg keys to mame defaults.
   >2. Every game you play with this on will reset to defaults when you exit.

Sorry, I must not be understanding it correctly. When I read the above lines it sounds like when you shut off legacy remapping your default.cfg file and all the <game>.cfg files will be reset?

> Just don't map the dpad in analog mode because that sits as a disabled axis in digital mode.

This line tells me if I map the dpad in analog mode those input values will be not be honored when I shut off legacy, so I would need to remap any dpad values that are used in analog with it off? Sorry I am just trying to wrap my head around how it works.

Wilstorm:
@grant2258 - I think I understand and I appreciate you explaining how it works. It can get complicated really quick. I just need to reread it a few more times. 

I just didn't want to turn it off and then back on and all my configs are wiped every time. I have a lot of custom configs. I setup my inputs so they work on both my IPAC and controller at the same time. So it will say keyboard key X or controller button A for jump, etc. Basically I use the same image on my arcade setup (without analog) as well as my controller (with analog).

I think having analog is great and you also, if I understand it correctly, seamlessly integrated analog into the Retroarch frontend.

I do know that everything you have implemented always turns out great so I'll test it out once it's ready to go into the latest build.

grant2258:
ill see how much code needs done for two config options.  I personally think its the only way round this that will keep all users happy.

Wilstorm:
@grant2258 - I think what you're saying will work because you can't assume the values in the TAB menu are what should be used in the game.cfg file when disabling legacy remapping. Are TAB values ignored when legacy is disabled? So you would need to fall back to the RA defaults. 

>Wiping out config files is a big extreme just a default and the game you are playing currently.

My other thought is about the game reset from the comment above. Yes it will reset the current game but it's already reset the default.cfg so how do you keep track of all the other games as you launch them without resetting all of them that may have incorrect values if that makes sense. TAB values can be something entirely different from the RA values.

Wilstorm:
I updated the above comment so hopefully it makes more sense.

grant2258:
what happens in the process is this if legacy mode is on save the config and get on with things.

if you using legacy mode off. It uses the defaults set in mame it just skips loading the default.cfg then  loads the game.cfg and sets everything back up that isint set to the defaults and save. It has to be done this way or you loose coin counters and dipswitches.

Wilstorm:
I understand what you're saying but for example you're playing Out Run and I disable legacy mapping. So it resets just Out Run to RA defaults and the default.cfg and saves.

Now what happens when I start Donkey Kong? Since it's a core option does it reset each game on launch if it see that legacy remapping was enabled but now disabled?

If it was always disabled it doesn't need to wipe out configs but if it was enabled and then disabled it might need to wipe the configs.

It just seems if you flip that option on or off you're going to make a huge mess of your configs potentially. 

>if you using legacy mode off. It uses the defaults set in mame it just skips loading the default.cfg

I don't quite understand above when you say it uses the defaults in MAME and skips loading the default.cfg. Isn't the default.cfg written by MAME so it is the default?

Wilstorm:
Another way to put it say for example you launch Donkey Kong, enable legacy remapping and setup your config and play a few games so now dkong.cfg is written out.

Then you launch Out Run, disable legacy remapping, setup your config play a few games and exit the game.

Now if you relaunch Donkey Kong again it has TAB defaults and that have been written in the dkong.cfg. How does the core know they are legacy enabled remapped values and should delete the dkong.cfg and write out a new one? 

The default.cfg has RA values but the dkong.cfg has TAB values so they could potentially be "orphaned" values and mismatch.

Ok the wife is harking to watch a movie. I appreciate explaining the new options even if it's a bit confusing for me.

grant2258:
yea found a buggy need work on it

UDb23:
As a general concept, agreeing with @Wilstorm, switching an option ON or OFF ideally should not wipe out any previous per game config. 
So in this case, if the coding effort required is reasonable, I'd go for the "two config" option.

btw: we will never thank you enough for all the hard work you're putting into mame2003plus.

grant2258:
Yea that was my original idea i said to him. Problem is mark wants to loose the tab menu eventually. The defualt configs work for most games ill have to think about this have a few little bugs going on anyway to smooth out.

Wilstorm:
I was thinking about this last night and one other potential pitfall is inheritance. Here's the way I always viewed the basic flow. Hopefully this diagram displays correctly.

  Core
     |
     |
default.cfg  --->  Input (general) from TAB menu
     |
     |
game.cfg  ---> Input (this game) from TAB menu

1) When you first launch a game it creates the default.cfg and game.cfg (i.e. default.cfg & dkong.cfg) if they don't exist.

2) When you launch a game, pull up the TAB menu and modify a setting in ```Input (general)``` it writes the change to default.cfg. 

3) Also when you launch a game the game.cfg (dkong.cfg) pulls updates/changes down from the default.cfg.

4) If you disable ```Legacy Remapping``` and reset the default.cfg then every game that has been launched
and has a game.cfg (dkong.cfg, outrun.cfg, digdug.cfg, etc.) will pull updates/changes down from the default.cfg. 

The exception being edits to ```Input (this game)``` wouldn't change the game.cfg (dkong.cfg) regardless of the state of ```Legacy Remapping``` being on or off because they are written to the game.cfg (dkong.cfg) directly.

Basically there's an inheritance hierarchy. If you ever used Group Policy in Active Directory or anything along those lines it's pretty much similar.

Wilstorm:
Sorry if that was a half baked thought. The thought being is the game.cfg files are linked to the default.cfg in "real time" for lack of better words. By that I mean if you change button 1 from say "g" to "h" in ```Input (general)``` and launch a game it will have the updated button 1 value immediately.

Basically any change to the default.cfg will have an immediate impact on the game.cfg files except for overrides through ```Input (this game)```.

The way it's designed makes sense that way you don't have to launch 1000 games to change 1 button. You only need to change the default.cfg at the root through ```Input (general)``` and let the change filter down the tree.

grant2258:
i just have a little fix to do for settings that arent default like f2 and the keys in space wars

grant2258:
ok the dual system was very easy to do. It will corrupt each other at the moment there is not flags to check what system you started with and changed too. 

The thing you need to keep in mind if you change the options to enable/disable  changing modes it wont save on exit. basically you need to restart when changing this option when i finish implementing it. you can still change controls ect but nothing will save until you next start.

grant2258:
Well I have a good overall picture of where I want this to go it has to be done in stages though. So im closing this for now and starting the updates that are needed right away

PR #635, opened 2019-01-28
stage1 updates ready to pull

grant2258:
*Fixed Digital and analog switching implementation

*fixed the cfg format to work with mame078 / mame2003 ( you need to delete you current +files)

*stopped none legacy mode saving config file

* fixed none legacy mode not setting some none standard mappings and analog stuff.

Stage two will make a dual saving system for legacy and none legacy so both will run separately and will make working on both system easier without corrupting each other.

A note when switching from digital to analog you will need to resetart for analog to work it will only work as digital because mame inits the digital on starting.


Wilstorm:
@grant2258 - Restart the core or restart the system when on a Pi 3?

Wilstorm:
@grant2258 - When switch to analog I mean do we restart just the core or the whole system on a Pi 3?

Wilstorm:
@grant2258 - Ok, sounds good, will do some testing and thank you. This whole piece sounds incredibly complex looking at all the commits you've been doing on this analog piece these past days.

grant2258:
all the commits where just testing different ways then see whats workable we will have a dual save system for both modes right now on only legacy saves

PR #638, opened 2019-01-30
avoid using "round" function (not in MSVC 2010)

markwkidd:
Howdy @grant2258 ! It seems MSVC's `math.h` simply doesn't have a `round` function. Our options as I know them include defining `round` behind an MSVC `ifdef` or avoiding using the round function.

This PR takes the latter approach -- and this should be right, right? But it's all up to you how to proceed.


grant2258:
you still using round why didi you change this ?

PR #644, opened 2019-02-01
finish the dual input system

grant2258:
Ok the input system is finished you can now switch between digital only and  digital and analog.

legacy and none legacy have there own config files so both will save dip switched and analog settings 

it you legacy mode you need to restart to enable saving dipswitch settings ect only if toggled.


robertvb83:
@grant2258 you removed dual joystick hack and automatically applie it for respective games in digital mode?
but what if someone want to play total carnage in 2 player mode? we need to switch off dual joystick hack to play with joystick and buttons or map shooting to joystick movement or whatever on a 2 player cabinet with only 2 joysticks total

grant2258:
you can change this in input this game if you need to map it differently

grant2258:
@robertvb83 let me explain it a bit easier if you swap to legacy mode and none legacy mode(restart required) they use there own settings.

legacy will be default dual joysticks

set none legacy mode tab this game to use the buttons and just swap modes for which one you want to use

PR #648, opened 2019-02-04
fix vectror gfx setting up

grant2258:
due to the change in aspect ratios that where requested to be change it had a knock on effect on vector games as the code was set to use mames aspect ratio .

they run 800x600 by default * whatever res you choose. since we clamped the aspect ratio for overlays this doesn't fair well it all been fix up and the resoultions are as follows 

x1 is 640 x 480
x2 is 1280 x 960
x3 is 1920 x 1140

@UDb23 this means you can do hd for vector games at x3 it will scale perfect.

the default settings are x2 with a beam width of 1 this can change by default if the pi coupes well with x3

### DIFF
--- a/src/mame.c
+++ b/src/mame.c
@@ -822,6 +822,8 @@ static void init_game_options(void)
   }
 
   /* update the vector width/height with defaults */
+  options.vector_width=640;
+  options.vector_width=480;
   if (options.vector_width  == 0) options.vector_width  = Machine->drv->screen_width;
   if (options.vector_height == 0) options.vector_height = Machine->drv->screen_height;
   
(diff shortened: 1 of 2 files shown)


UDb23:
Good news !
Thanks a lot, really appreciated.

src/mame.c
```diff
@@ -822,6 +822,8 @@ static void init_game_options(void)
   }
 
   /* update the vector width/height with defaults */
+  options.vector_width=640;
+  options.vector_width=480;
```
hhromic: this doesn't look right..  did you mean `options.vector_height = 480` here ? :)

grant2258:
oh dear ill fix it blonde moment thanks

grant2258:
lesson to myself one should never copy and paste twice!  lol

arcadez:
Measure twice cut once :)

grant2258:
haha class woodwork 101

PR #660, opened 2019-02-14
fix itech32 64bit segfault by updating video code

grant2258:
pretty sure drivers edge can be made playable with this updated code ill look at it another time. The inputs do need fixed up for a start


arcadez:
It ran like a dog on the Xbox so the Rpi might have trouble with Drivers Edge but i guess we wont
know till it's up and running

grant2258:
well might not be worth doing if its going to run like a bag of nails

arcadez:
Yeah i have the Chase Bombers code which would have been easy to port across but held off
from doing so due to performance likely being an issue on the many systems which use this
core i think Drivers Edge would be in the same ballpark.

cbombersj

PR #669, opened 2019-02-19
fix tekken2 segfault

grant2258:


grant2258:
@markwkidd can you please revert this teh cpu core is having 64 bit issues as well probably related to that thanks

PR #670, opened 2019-02-20
Revert "fix tekken2 segfault"

grant2258:
Reverts libretro/mame2003-plus-libretro#669

### DIFF
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -250,7 +250,7 @@ int osd_create_display(
       if (!video_buffer)
          return 1;
    }
-  if( video_conversion_type == VCT_PALTO565 )   video_buffer = malloc(video_config.width * video_config.height * video_stride_out);
+
    return 0;
 }
 


grant2258:
cheers big yean might look at this when i have a lot of free time dont know if its worth the effort though considerning there is no sound either

arcadez:
Nah forgot aboot it big chap it's no worth the hastle, on a related matter eg games where the
performance on most platforms that use this core would likely mean little payoff vs the work 
required i was looking at zn.c PSX hardware more or less and they're all working with sound
by the time MAME84 rolls around.

Might have been a viable target as the CPU's dont need touched just the driver, vid plus new
protection machine file and one sound core, we'd need a mirror or two though to get around
AM_SHARE in the mem map and after the hastle with galaga.c driver due to this maybe it's
best to leave it be :)

grant2258:
Ill try patch the video code up at some point that should be enough to stop the random ones segfaulting

PR #687, opened 2019-04-03
updates

grant2258:
Update dat hasnt been done in a while.

rename control mapping to and and digital to be more clear for users using digital only controllers.

also enable legacy mapping by default since Ra cant configure analog controls as we have a dual input system now it easy enough to switch between the two. 

RA is support is no where near ready to set mame controls up. The tab menu is a better option until someone fixes it up. I really dont see how that  will be possible since you cant have a user interface wimp interface in a core in ra.


markwkidd:
I'll find time to develop an alternative to legacy mapping on by default as
it exists now, but I'll not make that part of this change.

markwkidd:
I will mention this again, but RetroArch is not the only libretro frontend.

grant2258:
Well again its up to you want to do with it im not fussed either way i just changed some defaults setting instead of answering the same questions on the forums. It was kind of fine this way for digital when things mapped default (not all do) again thngs are overly complicated now for new users

grant2258:
@markwkidd i was going to get rid of the analog/digital setting but will wait till you finished what you are going to do and work with that too many options for users to understand at the moment. It going to require changes to the mame input descriptors

PR #690, opened 2019-04-07
tweak how timing is dealt with

grant2258:
It seems i have go to the bottom of what RA is doing to the timing there is no need to change the sample rate just make these games run at 60fps like RA wants mame just wont render the extra frames.

RA believes a framerate form 57 to 59 should be 60 by design. This option will be disabled by default if users preference is not playing a game to fast and high pitched audio they can just put this option on. 

https://github.com/libretro/RetroArch/issues/8555


grant2258:
actually im closing this till im compleley happy with what we are going to do

PR #766, opened 2020-01-03
Cyclone and DrZ80 support

### LINKED ISSUE #765: Cyclone and DrZ80
Can those commits:
https://github.com/libretro/mame2000-libretro/commit/ad07ac38747bd8abc74ea143dd73b7a05ee4d3be
https://github.com/libretro/mame2000-libretro/commit/1e09c86764f74b2a0ebad601df745cb7995d5c7f

Be ported to the mame2003-plus core? Can they be enabled for Pi0 as well?

frangarcj:
Fixes #765


twinaphex:
@arcadez2003 I think this would be good to merge.

arcadez2003:
Alrighty then

src/mame2003/mame2003.c
```diff
@@ -798,6 +801,101 @@ bool retro_load_game(const struct retro_game_info *game)
...
+#endif
+
+#if (HAS_DRZ80)
+	/* Replace Z80 by DRZ80 */
+	if (use_drz80)
+	{
+		for (i=0;i<MAX_CPU;i++)
+		{
+			unsigned int *type=(unsigned int *)&(Machine->drv->cpu[i].cpu_type);
+			if (type==CPU_Z80)
```
wn2000, line 873: This seems to be a bug. It should be `if (*type == CPU_Z80)` (gcc does warn about comparison between a pointer and an int). So DRZ80 was never actually enabled because of this.
But after fixing this (as well as on line 887), I lost part of the sounds in games (e.g. `sdfight`), until DRZ80 is turned off.
@frangarcj hope you are still around to take a look.

mahoneyt944:
Are you testing this with the current core? Our core now have core options to force enable these modes.

PR #769, opened 2020-01-18
Fix -mthumb for cyclone and drz80

frangarcj:
This allow to use mthumb with this asm cores


arcadez2003:
@twinaphex might be time to give this fella push access at least then he can send fixes
for the cyclone and drz80 code as and when he pleases.

Regards

PR #840, opened 2020-07-29
Update bootstrap.c

mahoneyt944:
update food fight bootstrap to include a joystick calibration for upright and cocktail. Tested this new nv bootstrap and it plays fine. Seems it's timing out Travis check falsely. May just be the nature of the file. Should be ok to merge unless someone knows otherwise?


mahoneyt944:
Tested this new nv bootstrap and it plays fine. Seems it's timing out Travis.

PR #859, opened 2020-08-12
Add pacman speedhack roms

jasoncmyers:
This enables the speed-up hack roms for Pac-Man, Ms. Pac-Man, and Jr. Pac-Man (pacmanf, mspacmnf, and jrpacmanf).

This is useful for people who want to be able to select between the fast and regular versions of these games without needing the players to access the system menu and dip switches.  It has the added benefit of tracking the high scores separately between the fast and regular versions.  

The fake "speed cheat" dip switch is left intact for the original roms, so it should not break any existing setups.  In the fast roms, the fake dip switch will still appear but it has no effect on gameplay; the speedup hack will be enabled regardless of its setting.

This works using the most recent versions of pacmanf, mspacmnf, and jrpacmanf, which have been unchanged since ~2010.  
Note that the jrpacman files have been renamed, but not changed, since the 2003 version, so only an unmerged jrpacmanf rom will currently work.  Making a merged set work would require either releasing a special 2003 version of jrpacmanf with files renamed to match 2003 jrpacman, updating to the new name scheme for jrpacman (which would break existing setups), or having some sort of dual-load functionality for the two different file name sets (which I'm not sure how to do).


markwkidd:
Historically we have not supported Merged sets, just Non Merged and Split.

This seem looks great!

Wilstorm:
@jasoncmyers - I don't quite know what happens in the background with the auto DAT generator mame2003-plus and if it would treat them as a parent/clone relationship. 

But either way in a merged, non-merged or a split set it shouldn't be an issue I wouldn't think as long as the ROM names are unique (even though they may have the same hash). This core loads based on hash value and not necessarily name. It would probably double up matching hash ROMs because they are named different I would imagine.

The bigger issue is same-name different hash ROMs in parent/clone relationships which are usually left up to the ROM management programs (like ClrMamePro) to handle those types of conflicts, depending on your settings.

I don't know the code well enough without seeing a test DAT of how it would look but I would think you would want to be cautious as to not break those classic games. It would cause some serious outcry I would imagine. :)

jasoncmyers:
Interesting, I hadn't looked into the core code enough to know about the loading by hash.  I just assumed that the filename changes between versions of jrpacman would break a split fast rom and didn't try to track down a split version to test.  But I found one this morning and sure enough, it works fine.  So, I guess disregard that final paragraph.

twinaphex:
Cool that you're still active @markwkidd , very good to see!

Wilstorm:
Yeah, exactly you can name the ROMs 1, 2, 3, etc. for all intents and purposes and it really doesn't matter.

In a non-merged sets it wouldn't matter but in a merged or split set scenario I would be curious if it adds the ROM twice or ups it to the parent (jrpacman/jrpacmanf) knowing the clone can find it there based on hash. I suppose it would be easy enough to build a small test DAT to find out.

The piece I don't know well enough is @grant2258 added an auto-generation feature that allows you to create a current on-the-fly DAT while in any game. I would guess these new additions would be treated as clones which would possibly make the question above viable but on the other hand the core could care less if a ROM is located in the clone or the parent. It's just more the idea of caution I guess.

The other piece is how @markwkidd wants to handle those type of additions. I know in the past they have tried to steer clear of to many hacks in the core.

Wilstorm:
I wish I could be of more help here but just to add a little more information. Merged sets don't work for this core. I would imagine it's a front-end feature (Retroarch) is why it doesn't work/recognize merged sets and splits out the individual games. That means you'll always need at least one ROM in the clone zip files in order for this core to recognize a game.

That leaves non-merged and split-merged. Both which do work fine. Non-merged isn't an issue so that just leaves split-merged.

It's a unique situation where you want to keep the original ROMs but also add a hacked set that will use the same ROMs named differently. Knowing that ROMs will load based on hash values helps things work but I think it boils down to how you want to organize the ROM sets based on the merge tag.

This is where Grant really knows how it works. I don't know if the DAT generator will try and match both name and hash of a ROM before setting a merge tag in the DAT or if it's hash based only. I would assume it would be both vs. hash only and trying to resolve naming conflicts on merge. 

If it uses both criteria then it will leave the duplicate ROMs in the parent/clone with different names and have a clean separation. If it's hash only then it will try and merge them up to the parent with either the clone or parent ROM name, which I don't know based off of the code. The game should still run either way but it leaves a bit of mess with ROM names in one of the scenarios.

If someone knows the code well enough to understand how the merge tags are applied during DAT creation would answer the question or I would say the simplest way is to just generate a DAT from the updated code and look.

markwkidd:
I appreciate that I still have merge permissions on this repository, but I don't on the other hand feel like I represent the priorities of the developers currently involved in the project.

Would it help for me to compile this branch, generate a DAT, and upload it here for consideration? I would like to make sure active contributors have a say in adding hacks even when, as in this case, it seems like a popular title/hack combination.

jasoncmyers:
I'm obviously pretty new to this, but is the DAT you're talking about the XML file generated through the MAME menu?  I just ran that and took a look.  I'm not entirely sure how to interpret all of it, but it looks like it recognizes the file names in the original jrpacman rom and marks them down as the `merge=` filename for jrpacmanf.

Example:
`<rom name="fast_jr.8d" size="8192" crc="461e8b57" sha1="42e25d384e653efb95a97bd64f55a8c3b3f71239" region="cpu1" offset="0"/>
		<rom name="jrp8e.8e" merge="jrp8e.bin" size="8192" crc="ec889e94" sha1="8294e9e79f8fd19a419431fa690e6ac4a1302f58" region="cpu1" offset="2000"/>
		<rom name="jrp8h.8h" merge="jrp8h.bin" size="8192" crc="35f1fc6e" sha1="b84b34560b9aae18b24274712b052283faa01730" region="cpu1" offset="8000"/>`

Hmmm, I can't seem to figure out how to make a newline inside that code block.  But hopefully that's readable enough.  I can also attach the whole file, if desired.  But it's 19 MB and apparently github doesn't like .xml extensions.

Wilstorm:
@markwkidd @jasoncmyers - Thanks for uploading the DAT as I learned some new things also. Looking at the DAT basically what it's doing with a split-merge (or merged) set is renaming the ROMs to the parent set names (old names) and discarding the clone names (new names). So the core will need to load the hacked versions based off the hash. If it's a non-merged set it will leave the ROM hack versions with the new names. It looks it should work fine without causing issues with existing setups and merge nicely with the existing ROMs, no need to keep an unmerged set separately.

Something of note. Out of curiosity I made a test DAT where the clone set had all matching hashes to the parent but different names. When doing a merged or split-merge set it discards the clone set zip. 

So for example if game1.zip and game2.zip have the exact same ROMs but named differently and you do a merge or split-merge set it will merge and rename all the roms into the game1.zip and delete game2.zip. Even though there are two games in game1.zip Retroarch would only see game1.zip and never know game2.zip exists in the same zip. That's basically how a merged set works. I don't think merged sets are a core feature but more of front-end feature.

The DAT generator is quite cleverly implemented! :)

@markwkidd - Is it easy to compile a branch? I basically use RetroPie and I'm quite shielded on how things work. I pick an option in the RetroPie menu and it compiles the current branch so I don't see what commands are running.

> Hmmm, I can't seem to figure out how to make a newline inside that code block.

@jasoncmyers - If you figure that out let me know! :)

markwkidd:
It's early to know for sure, but I believe this pr was a victim of the libretro hack. That has no bearing on the usefulness of the pr -- it may need to be recreated though

markwkidd:
@jasoncmyers this PR was forced permanently closed as a result of the recent libretro hack. Could you please submit a new PR with this patch? I apologize for the inconvenience

PR #870, opened 2020-08-20
Add Pac-Man speed-up roms

jasoncmyers:
Re-opening this as requested by @markwkidd:

This enables the speed-up hack roms for Pac-Man, Ms. Pac-Man, and Jr. Pac-Man.

This is useful for people who want to be able to select between the fast and regular versions of these games without needing the players to access the system menu and dip switches (e.g., cabinet builds). It has the added benefit of tracking the high scores separately between the fast and regular versions.

The fake "speed cheat" dip switch is left intact for the original roms, so it should not break any existing setups. In the fast roms, the fake dip switch will still appear but it has no effect on gameplay; the speedup hack will be enabled regardless of its setting.

This works using the most recent versions of `pacmanf` and `mspacmnf`, which appear to be unchanged since their initial 2004 inclusion.  The Jr. Pac-Man fast rom was released later, when different naming conventions were being used, so I renamed it from `jrpacmanf` to `jrpacmnf` to remain consistent with the 8-character limit used in all the 2003-plus sets I've seen.

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -126,7 +126,7 @@ const struct GameDriver *test_drivers[] =
 	DRIVER( vanvan )	/* (c) 1983 Sanritsu */
 	DRIVER( vanvank )	/* (c) 1983 Karateco (bootleg?) */
 	DRIVER( alibaba )	/* (c) 1982 Sega */
-	DRIVER( alibabab )      /* (c) 1982 Sega (bootleg) */
+	DRIVER( alibabab )	/* (c) 1982 Sega (bootleg) */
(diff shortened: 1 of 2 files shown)


mahoneyt944:
@twinaphex I don't have merge access currently.

markwkidd:
@twinaphex I don't have merge access any more as of the hack. I would be interested in having those permissions again -- I try to keep somewhat up to date with github and it has been handy for me to be able to merge someone's PR once in a while even though I'm not doing much coding now

PR #873, opened 2020-08-24
Uni-bios_4_0.rom support NEOGEO

mahoneyt944:
Uni-bios_4_0.rom for neogeo now supported


mahoneyt944:
@twinaphex could I get access? Thank you. @markwkidd was also looking to get access as well.

twinaphex:
Sure, my bad if permissions were lost after the hack.

PR #878, opened 2020-09-08
TEST DON'T MERGE

mahoneyt944:
nbajam, reviewing drivers.


mahoneyt944:
@arcadez2003 Looking at advmame, the williams_sndhrdw and tunit stuff have many changes that would probably eliminate the sample we use in our core? At a quick glance these changes seem feasible.

mahoneyt944:
Maybe we should change this to default disabled?

PR #879, opened 2020-09-10
Add support for Mortal Kombat 2 + and Ultimate Mortal Kombat 3 +

Zoltan45:


mahoneyt944:
@arcadez2003  This PR looks like it could use the color pallet fix up you recently added. Otherwise I think it's safe to add.

arcadez2003:
@mahoneyt944 You dont need to run anything by me :) if ya wanna make changes go ahead the only rule we
have here is ya break something ya fix it, which unfortunately some people forget and leave it up to others to sort
midway samples spring to mind.

Which reminds me after speaking with @barbudreadmon i'll need to check if the cyclone CPU cores which were
added to this project a while back are turned on by default as these cores can be problamatic and actually
break more than they're proposed to fix.

Ideally in this core i'd like the cyclone CPU's to be an option which can be turned on for the very small userbase
who still use an Rpi 1 and 2 and would require the speedups in certain games which the cyclone brings to the table
but i dont want it enabled by default as the vast majority of users do not require it.

And certainly even more so if games are broken in the process.

mahoneyt944:
@arcadez2003 yeah I just merged a fix for the driver invasion of ost. Cool option but could have been implemented a bit smoother. Should be better now. For this PR though I just wanted to see if I should add the pallet fix you recently added to the other mk3 titles. I think this would be needed here too from what I can tell. But I don't know much about what that fix did and if it's needed here for sure

mahoneyt944:
@Zoltan45 I did some cleanups, but it looks like you are defining speed up 4 but using speed up 3 in the init for umk3. I would also review the umk3_palette hack that was done to see if it applies or not.

Zoltan45:
@mahoneyt944 I thought I cleaned everything up, something got messed up in my final commit from my pc to github i think. I will go over it this evening (AM here currently).  Thanks for giving it the once over.

mahoneyt944:
@Zoltan45  Which version of plus is this? I see on their site they have different betas. The crc and sha1 needs to match the version you are using.

mahoneyt944:
I merged this and made some format tweaks. Tested and both are working. I will be fixing the low volume initiation and CMOS errors soon, which is a midway issue not related to this merge. Thank you for you contribution. This is a great addition to the core.

PR #902, opened 2020-10-12
Vball color fix pal4bit

mahoneyt944:
Color correction.


mahoneyt944:
@arcadeez now that we support bit color conversations. I'm thinking about sweeping through the video drivers and updating as many as possible like this one. There's obvious improvements in the color and picture quality and I don't think it will impact performance much at all. What are your thoughts on it?

PR #905, opened 2020-10-22
dodgeman wip 2

mahoneyt944:
@arcadez2003 tagging you in case you have any input on this. thanks. 

- ay2 still not hooked up properly yet. not sure how to differentiate ay1 from ay2 in machine start and the port write 26 and 27.
- graphical glitches on logo screen and with craters in game. might be due to the driver init hack being used?
- still using irq hack `(irq0_line_pulse,8)`, battlex_interrupt not hooked up. the hack seems to work so if we can get dodgeman to work with it too might not need to do this one.
- need to finish updating driver_init to use `int tile_size = memregion("gfx1")->bytes() / 32;` instead of hack (0x1000 / 8). needs ported. sizeof?
dodgeman-201022-101339
dodgeman-201022-101300
dodgeman-201022-104701


arcadez2003:
@mahoneyt944 Try this i've hooked up the second AY updated the graphical handling to match up with how it was
when dodgeman was added, i've not tested this whatsoever so dont know if the sound is correct nor if the gfx update
in the driver will line up with the video code in battlex_vidhrdw without that needing updated also.

[stuff.zip](https://github.com/libretro/mame2003-plus-libretro/files/5425343/stuff.zip)

mahoneyt944:
@arcadez2003 memory_region_length was just what i needed! thanks. graphics fixed.
dodgeman-201022-173543
dodgeman-201022-173613

mahoneyt944:
updated:

- ay2 still not hooked up properly yet. not sure how to differentiate ay1 from ay2 in machine start and the port write 26 and 27.
- still using irq hack (irq0_line_pulse,8), seems to work just fine without hooking the interrupt. might just ignore this.

PR #908, opened 2020-10-29
IGMO color prom

arnoldsecret:
Untested.
[0.223: Andrew Welburn and Craig Anstett added color prom to IGMO, palette is now correct. Fixed rom names.]


mahoneyt944:
tested - confirmed

PR #922, opened 2020-11-15
Dottori-Man Jr.

arnoldsecret:
Untested, but the dotrikun driver's simple enough to be an easy addition. Interesting hack on a test board that looked like this

http://adb.arcadeitalia.net/media/mame.current/pcbs/dotrikun.png?release=191


mahoneyt944:
Can't see any harm in this. Will have to test after merging.

PR #924, opened 2020-11-20
kof game add

odarknighto:
The new game source code is from mame103,104,105


odarknighto:
The new game source code is from mame103,104,105

arnoldsecret:
Question, have you tested KOF 2003, changes in decryption code for neogeo in later versions might break what we have right now.

odarknighto:
> Question, have you tested KOF 2003, changes in decryption code for neogeo in later versions might break what we have right now.



> Question, have you tested KOF 2003, changes in decryption code for neogeo in later versions might break what we have right now.

I had test kof2003 and work fine,I just add so I haven't change kof2003d init,and kof2002 I have change Decrypt way like mame103 and work fine

odarknighto:
By the way,I only test with my Nintendo switch because compiler easy than other platforms,sorry><

arcadez2003:
@odarknighto if you've tested the changes and there is no game breakage with regards to kof2002 / kof2003 / kof10th
and there are no compile issues then we can push it to mainline thanks for your contribution

It'll be good to actually support the proper version of kof2003 in this core i didn't think it was doable due to the 
aforementioned changes to the neogeo decryption circa MAME 103 i guess i was wrong :)

odarknighto:
> @odarknighto if you've tested the changes and there is no game breakage with regards to kof2002 / kof2003 / kof10th
> and there are no compile issues then we can push it to mainline thanks for your contribution
> 
> It'll be good to actually support the proper version of kof2003 in this core i didn't think it was doable due to the
> aforementioned changes to the neogeo decryption circa MAME 103 i guess i was wrong :)

actually kof2003 I used mame104 source because p3-decrypt  is difficult to find and mame104 can use encrypt p3 and play well
Those add game I had try and can play😀

odarknighto:
> @odarknighto if you've tested the changes and there is no game breakage with regards to kof2002 / kof2003 / kof10th
> and there are no compile issues then we can push it to mainline thanks for your contribution
> 
> It'll be good to actually support the proper version of kof2003 in this core i didn't think it was doable due to the
> aforementioned changes to the neogeo decryption circa MAME 103 i guess i was wrong :)

I don't know other game because I just try add kof game,but I can promise I haven't edit other game driver ><

odarknighto:
> @odarknighto if you've tested the changes and there is no game breakage with regards to kof2002 / kof2003 / kof10th
> and there are no compile issues then we can push it to mainline thanks for your contribution
> 
> It'll be good to actually support the proper version of kof2003 in this core i didn't think it was doable due to the
> aforementioned changes to the neogeo decryption circa MAME 103 i guess i was wrong :)

About neogeo decrypt I haven't change anything so it's ok

PR #925, opened 2020-11-22
berabohm inputs

arnoldsecret:
0.89: Nicola Salmoria fixed inputs and DAC sounds in Beraboh Man, two sounds at the same time were causing severe slowdowns. Added 'Invulnerability' and 5x 'Unknown' dipswitches.


arcadez2003:
Port tagging is not compatible with MAME78............

For example.......

PORT_START_TAG( "IN0" )

It needs to be.......

PORT_START

arcadez2003:
Alright the input ports should be fine now, however looking at the driver you picked the wrong build
to backport from as it comes slap bang in the middle of a huge update to the Namco System 1 code and as such
it wont be easy to isolate what code you require from all the other changes which were done.

https://github.com/mamedev/historic-mame/commit/51ddfdc82c9f7ffef66691717265b4405469d0ab#diff-aa9f24f35c3ec8140dbf6364302e45f095b4b997f72f2df670e29f8063400d33

There are some alterations to the DAC code which you didn't backport i suspect without those you'll see no real benefit
by just updating the inputs, personally i'd start by studying the code changes with regards to the DAC sounds
and check the sound cpu memory map to see if any address values were changed as part of that.

Then i'd backport the code into our driver and then test all the games to make sure there is no sound breakage
in any of them due to changes i might have made, then check the speed in berabohm to see if i got the performance
boost im looking for.

Only after all of the above would i push to the git.

arnoldsecret:
did a couple commits in my fork regarding the sound, needs some testing
https://github.com/arnoldsecret/mame2003-plus-libretro/commit/0e5abc61b9cf574a0636e935009e731a9147b665
https://github.com/arnoldsecret/mame2003-plus-libretro/commit/104fcdfd4b193d410c1b17b71add838da727ddf4

PR #930, opened 2020-12-04
add neogeo m1 decrypted program

odarknighto:
add decrypted program from mame128.
Now all m1 rom can use encrypted rom to play and no need m1_decrypted!
(game kof2000,kof2000n,kof2001,kof2001h,kof2002,kf2k2pls,kf2kpla,kf2k2mp,kf2k2mp2,kof2003,mslug4,rotd,matrim,pnyaa,samsho5,samsh5sp,mslug5 are change to use decrypted)


arnoldsecret:
Impressive, though I'm concerned about performance on weaker systems (pi3, original xbox)

arcadez2003:
> 
> 
> Impressive, though I'm concerned about performance on weaker systems (pi3, original xbox, android)

Performance should not be an issue really, as we're only decrypting the sound rom on boot rather than using
a pre decrypted one beforehand, anyway some nice updates here it'll make obtaining the correct roms for certain
NeoGeo sets far easier.

@odarknighto im happy to push this to mainline if you have tested every game these changes affect and can
corfirm they're all still playable.??

odarknighto:
> Impressive, though I'm concerned about performance on weaker systems (pi3, original xbox, android)

I just try switch so I don't know the performance,but mame2010 also use this on android so I think it ok!

odarknighto:
> > Impressive, though I'm concerned about performance on weaker systems (pi3, original xbox, android)
> 
> Performance should not be an issue really, as we're only decrypting the sound rom on boot rather than using
> a pre decrypted one beforehand, anyway some nice updates here it'll make obtaining the correct roms for certain
> Neo Geo sets far easier going forward.
> 
> @odarknighto im happy to push this to mainline if you have tested every game these changes affect and can
> corfirm they're all still playable.??
> 
> Edit you could have saved yourself a little trouble by just making audiocrypt into REGION_USER2 but no matter
> it's not how you travel to the destination getting there is the main thing :)
> 
> ROM_REGION( 0x80000, REGION_USER2, 0 ) \
> 
> void neogeo_cmc50_m1_decrypt( void )
> {
> UINT8* rom = memory_region( REGION_USER2 );

I don't  sure about this because in mame128 label 'audio' = REGION_USER2 ,audiocrypt is new label for decrypt m1,so change this I don't sure this ok!
But I have test all m1 rom and it's run ok in switch!!
Next time I maybe change all decrypt programs from neogeo.c to neocrypt,and let some decrypt programs (ex:kof99) change structure to same with mame103 for other people can edit easily!

PR #956, opened 2020-12-30
Imago

mahoneyt944:
Working now


mahoneyt944:
@arcadez2003  works now, but galaxian_draw_stars is commented out at the top and bottom of the vidhrdw. Kept getting undefined error for it. Probably a syntax thing? Do you see anything?

PR #977, opened 2021-02-10
Single default mapping

mahoneyt944:
- Eliminates analog / digital core option
- single default mapping
- officially support for middle mouse click
- map 6 players by default
- set UI select and cancel buttons to A and B for classic layout
- set pedals to L and R buttons for classic layout.
- fix analog scaling


Wilstorm:
@mahoneyt944 - Just a quick question to clarify. I usually do Libretro default mappings on my controllers. So on an XBOX controller for example controller Button B is RetroPad A and controller Button A is RetroPad B. It's been a while but I believe it would switch the A and B button for you so it's correct in the core. 

Does the UI select and cancel buttons get switched now with this update so using the Libretro defaults is going to switch them backwards. I hope that makes sense. Here's a snapshot of what I talking about for defaults. It's an XBOX controller and not a 360 but works for illustrative issues.

7f50121a-42c0-11e6-8ef5-50c4b5672216

mahoneyt944:
@Wilstorm we follow the retropad standard as also shown in the retropie documents for setting up your controller.

https://retropie.org.uk/docs/Controller-Configuration/

This means our ui buttons will be the retropad A and B, Which labeled opposite on the xbox controller.

PR #995, opened 2021-03-24
improve mame2003.h comments

markwkidd:
Before I go through `mame2003.c` and attempt this kind of process at larger scale I wanted to submit this PR first.

If this makes sense, I would follow up with another that adds function headers in `mame2003.c` but also comments within functions in areas related to input to the extent possible.

### DIFF
--- a/src/mame2003/mame2003.h
+++ b/src/mame2003/mame2003.h
@@ -249,28 +249,6 @@ void osd_stop_audio_stream(void);
 
 /******************************************************************************
 
-	Timing
-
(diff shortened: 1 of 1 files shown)


mahoneyt944:
You should probably create a new branch within this repository. On every commit to that branch you can download the built binary from the pipeline for testing purposes. The pipeline builds all branches in this repository. Then you can PR the branch.

markwkidd:
That sounds good. I'm closing this and moving to a branch.

PR #997, opened 2021-03-26
improve comments and clarify input code

markwkidd:


markwkidd:
There is still more commenting/documentation that can be done, and I'll probably do a bit more, but I would like to squash and merge this PR soon and start another. I'll check in a bit to make sure gitlab is building the PR OK before I merge it though! :)

markwkidd:
OK, having passed most builds by this point, and with no changes to operational code, I'm going to squash this into one commit and merge

PR #998, opened 2021-03-28
Updategames202103

jamiecurmi:
Hi guys! Added sound to a number of existing games, and added 3 new games (and some bootlegs)

Full details:

https://github.com/jamiecurmi/mame2003-plus-libretro/wiki/Update-March-2021

I hope you are all safe and healthy. Jamie


mahoneyt944:
Galactic and spacmiss need better rom naming as to not cause conflicts in clrmamepro. Having multiple rom names with different crc values. Where did you get these rom names from?

jamiecurmi:
Hi @mahoneyt944. This seems to be the naming used on the distributions around the net of the ROMs, and also used in MAME on GitHub. However, I also agree the naming leaves a lot to be desired, I just went with what was being used out there. As I'm new to this, so not sure what the approach should be.

I mean, we could rename all those in galactic to galactic_1.bin, galactic_2.bin etc., and spacmiss uses the same rom names for those with the same CRC, except for spacmiss_1.bin and spacmiss_8.bin being renames? Or all the ones in spacmiss could be spacmiss_N.bin for all (N = 1..8)? It just means if downloading the roms elsewhere, people will need to rename?

Let me know what approach you would like, or if there is some other approach? Thanks @mahoneyt944. Jamie

mahoneyt944:
Our roms load by their crc and sha1 checks first then use the name if it wasn't found by those checks. So I don't think renaming them would be an issue as long as the crc and sha1 checks are proper. We typically would name the parent set, galactic, then use those names for the spacmiss set were appropriate. 

As far as the 8 character naming convention... That is what we typically use but not sure it would be an issue either way. 

@arcadez2003 what would you suggest?

jamiecurmi:
I will make those suggested changes @mahoneyt944, and retest to make sure nothing breaks. I think for consistency it would be better if I renamed the 9 character names to 8, so I will also do that. I'll then make a new pull request if that works for you guys.

mahoneyt944:
If you commit to your branch(jamiecurmi:updategames202103) it will update here as well

jamiecurmi:
I've updated the files. Tested and all seem to be working fine. Please let me know if you have any other concerns, or it wasn't quite what you were expecting.

mahoneyt944:
@jamiecurmi you'll have to add these fixups to the changelog as well https://github.com/libretro/mame2003-plus-libretro/blob/master/CHANGELOG.md

Under "New Games", "New Clones", and "Games now with sound" as appropriate.

jamiecurmi:
@mahoneyt944, done. I've committed to this branch. If you can't merge from this and want me to put into a new branch with a new pull request let me know. I've also added changes for my previous 2 pull requests that were accepted and merged before this one – my apologies that I didn't realise I should maintain the CHANGELOG.md; I will do so in the future.

mahoneyt944:
I copied over your change log update.

PR #1007, opened 2021-04-03
small comments addition & cleanup

markwkidd:


mahoneyt944:
@markwkidd I'm not sure why lgtm.com is bugging out. Maybe I worked it too hard ha. Know anything about this check? I'm not sure we really need it at all now that the pipeline builds live.

markwkidd:
lgtm does more of an analysis where it can identify things that compile but might not be intended, etc. that being said, I don't feel too concerned about missing it on this particular commit. maybe this repo did hit a limit for its use :)

PR #1012, opened 2021-04-05
declare all-digital joypad

markwkidd:
The purpose is to satisfy the need to always have an all-digital retropad as the baseline


mahoneyt944:
Curious though, do we need an all digital base line? Could all digital be satisfied by our analog retropad essentially not using all the inputs?

markwkidd:
I would like it better if we didn't have to declare a digital joypad, and it seems to me like it shouldn't have to be this way.

I have a memory that this is an intentional RetroArch behavior based on their interpretation of the libretro API. However I am looking and I can't find anywhere this is explained that way. I'm still looking...

mahoneyt944:
Maybe ask on discord it seems odd to me too.

markwkidd:
Originally, I tried to make the Classic layout a `SUBCLASS` of `RETRO_DEVICE_JOYPAD` but when I did that, a generic RetroPad also appeared in the list.

The only way I got rid of it was to make one of our layouts (Classic) use the JOYPAD code, which "renamed" RetroPad in the menu.

markwkidd:
Yeah I may pop into discord for this one, good idea

mahoneyt944:
Alright, yeah because this also breaks the controls again :(

markwkidd:
I will need to do a little more work on this PR, probably an issue with it in the libretro control description functions

markwkidd:
There is a RetroArch issue open about this: https://github.com/libretro/RetroArch/issues/6966

Hmm.

mahoneyt944:
Maybe just revert that commit then. That might be why it was that was to begin with

markwkidd:
Yeah I'll revert in a moment.  Annoying that we can't do much about a
frontend issue

PR #1017, opened 2021-04-06
Lightgun v3 WIP

markwkidd:
Trying this again with the latest refactoring as a base


markwkidd:
It's working! Generally speaking 🤣 

In the latest commit, there is odd behavior when the lightgun coordinates pass off the screen. Probably need a couple of `if` statements to check at the boundaries.

Also I am having trouble getting RetroArch to accept control descriptions for the lightgun's controls so that they appear right in the RetroArch Quick Menu. I think my implementation may be wrong, but I am also willing to consider that there is a bug in RA.

PR #1018, opened 2021-04-06
delete

markwkidd:
once more...

Currently a bit buggy:
* description of controls for libretro
* when the lightgun gets near the edge of the screen area
* simulated offscreen reload is not hooked up between libretro and MAME (may be tricky)
* more to come...


markwkidd:
hmm, I can't seem to

PR #1028, opened 2021-04-12
lightgun api implementation

markwkidd:


markwkidd:
This PR implements a new libretro input device "Lightgun + Keyboard Encoder", selectable in the frontend. When this device is selected, MAME does not map or poll the mouse and retropad. The gun is not polled in any other input mode.

@mahoneyt944 this does not suffer from the MAME remapping issue that the old PR did. If you are willing, I'd like to add this as a solid next step before going further.

Eventually I would also like to try to have an 'everything at once' mode with RetroPad, Keyboard, Mouse, and Lightgun, but that would take even more refactoring, in MAME functions I am not as familiar with :D

mahoneyt944:
@markwkidd it looks ok to me. One thing though, I'm not sure we need these lines since we blank everything initially anyhow:
```
retroJsState[port][OSD_MOUSE_BUTTON_2]  = 0; /* no right click for a pointer */
retroJsState[port][OSD_MOUSE_BUTTON_3]  = 0; /* no middle click for a pointer */
```

markwkidd:
you're right about that - no need to blank those fields a second time. I just pushed a commit to remove those lines

mahoneyt944:
I think it's good to merge

markwkidd:
thanks!

PR #1037, opened 2021-04-22
latency & clarity: move libretro polling closer to MAME polling

markwkidd:
A simple idea that occurred to me while refactoring input code


markwkidd:
Any suggestions on a good game to use to see if this actually reduces latency? I suppose it should be the 'pause and frame advance' method.

mahoneyt944:
Start with the master branch then add these changed to it to test. The commit you merged here was old.

PR #1043, opened 2021-05-04
add comments and better labels for existing analog axes input polling

markwkidd:
no-op


markwkidd:
INPUT_BUTTON_AXIS_THRESHOLD is the RetroArch terminology FWIW

PR #1044, opened 2021-05-04
WIP refactor input caching - second attempt

markwkidd:


mahoneyt944:
breaks mame menu. inputs going crazy

markwkidd:
On the upside, I do have a controller with analog. I understand how the
existing code works. This analog to digital code may be something we can
let the frontend handle in the future, but for the moment I'm just trying
to cleanly refactor the current situation.

mahoneyt944:
I try to pop in to help test since this has a major impact on the end users experience.

Though we might want to let some time pass in-between these big overhauls to the input code to avoid the needle in the haystack debacle. Allowing time for serious testing in-between.

markwkidd:
I appreciate it. I have a complex schedule so I'm trying to find focused
time as I can 🤸

PR #1045, opened 2021-05-05
WIP for discussion: let frontend simulate digital dpad on analog joysticks?

markwkidd:
Historical context: When this core was first ported, I was not involved, but I know there was a less of an libretro API to work with. There were fewer, and simpler, callbacks available for input. That simpler libretro API is reflected in the way mouse/pointer input was coded, lightgun, etc.

Today the API and reference frontend are more sophisticated.

One of the areas where I assume frontends were less sophisticated back in the day is simulating digital button presses for analog joysticks. When the core was originally ported, I'm not sure there was even analog input support in RetroArch. 👯 

I wonder: do we need to implement analog to digital in the core any more, or can we outsource digital button simulation to the frontend?

When I use this branch, and I go to RetroArch's input menu and set the Analog to Digital option to Left, things seem to work the same. I have a simulated dpad on my analog joystick.

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -1269,18 +1269,6 @@ void retro_run (void)
     analogjoy[port][2] = analog_deadzone_rescale( input_cb(port, RETRO_DEVICE_ANALOG, RETRO_DEVICE_INDEX_ANALOG_RIGHT, RETRO_DEVICE_ID_ANALOG_X) );
     analogjoy[port][3] = analog_deadzone_rescale( input_cb(port, RETRO_DEVICE_ANALOG, RETRO_DEVICE_INDEX_ANALOG_RIGHT, RETRO_DEVICE_ID_ANALOG_Y) );
 
-    /* Analog joystick - read as digital button /*
-    /* If the analog value (normalized for MAME to the range -128, 128) is greater in absolute   */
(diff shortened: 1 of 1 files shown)


mahoneyt944:
Using the analog to digital option creates a double input. This option basically sends a digital signal along side the analog one which messes up our mame mapping since it detects both signals.

markwkidd:
Do you know of a good way to trigger the double input? Here's what I'm trying to do now to see it:

* Use this branch
* Set RetroArch "Analog to Digital" to "Left"
* Load sf2.zip
* Open MAME menu to controls for current game
* Select Up. Press up on the Left analog joystick.

When I do this, the MAME remapper maps RP1 HAT Up as I would expect, but I don't the double input.

mahoneyt944:
Well yeah if you remove the analog mapping then you don't get a double input. You just can't properly map the analog anymore. We want to support both individually

markwkidd:
Ah, OK I had tested to make sure that analog input still works (with sinistar.zip), but I see that the analog inputs can't be remapped in the MAME remapper with this code.

Now I have a test case to understand this. It does help to have a controller with analog input.

mahoneyt944:
Yeah analog inputs get messy because we want them to work as analog when appropriate and to simulate digital when analog isn't being used so things just work out of the box. Analog to digital was also suppose to get a update to not send the analog signal at all once switched on.

The only reason to use analog to digital is if the core does not natively support analog allowing the user to use it as digital.

PR #1051, opened 2021-05-20
add helper to pre-set input sequences with 7 codes

markwkidd:
I'm trying to break down some larger changes into small PRs. This one is pretty simple -- it just adds helpers to let us pre-set input sequences with up to 7 codes (for example: [keyboard up OR dpad up OR left joystick up OR lightgun dpad UP]).

This functionality already exists in terms of the core being able to handle 16 codes per sequence, there just aren't functions to pre-set that many codes until now.

### DIFF
--- a/src/drivers/pacman.c
+++ b/src/drivers/pacman.c
@@ -775,11 +775,17 @@ static MEMORY_WRITE_START( mspacman_writemem )
 MEMORY_END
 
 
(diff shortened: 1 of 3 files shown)


markwkidd:
how in the world have I pulled in a ms pacman commit into this PR. remaking...

PR #1052, opened 2021-05-20
add helpers to set 7-code input sequences

markwkidd:
I'm trying to break down some larger changes into small PRs. This one is pretty simple -- it just adds helpers to let us pre-set input sequences with up to 7 codes (for example: [keyboard up OR dpad up OR left joystick up OR lightgun dpad UP]).

This functionality already exists in terms of the core being able to handle 16 codes per sequence, there just aren't functions to pre-set that many codes until now.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -588,6 +588,33 @@ void seq_set_5(InputSeq* a, InputCode code1, InputCode code2, InputCode code3, I
 		(*a)[j] = CODE_NONE;
 }
 
+void seq_set_6(InputSeq* a, InputCode code1, InputCode code2, InputCode code3, InputCode code4, InputCode code5, InputCode code6)
+{
+	int j;
+	(*a)[0] = code1;
+	(*a)[1] = code2;
+	(*a)[2] = code3;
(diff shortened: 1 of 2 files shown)


mahoneyt944:
So we defined SEQ_DEF_6, but never had a seq set for it? That seems odd.

markwkidd:
I don't think any of the even-numbered sequence helpers get used -- a quick search is not showing anywhere that the existing SEQ_DEF_2 or SEQ_DEF_4 macros being used either.

mahoneyt944:
Feel free to merge whatever / whenever, don't wait on me ha. Though I'm always happy to help test code if you need.

markwkidd:
Thank you, I do greatly appreciate a second set of eyes on my code :)

I honestly have been thinking that there really is no purpose to setting a sequence with an even number of codes (2 or 4 or 6), but also that it doesn't harm anything to maintain the pattern and implement the even numbers. I don't need to get tempted to refactor MAME code just to make it cleaner in other words!

PR #1055, opened 2021-05-24
add and use codes for lightgun inputs, enable players 7 and 8

markwkidd:
This PR adds a new range of input codes for the lightgun and @mahoneyt944 this is another time when your review would be welcome. I believe this is working.

The value in this PR is to allow the same MAME control to be mapped simultaneously to a keyboard key, retropad button, mouse button, and a lightgun, which is what I do here for the common mappings.

If anyone does want to test, what I have done is mapped the RetroArch lightgun to my mouse, then used the MAME 2003+ core option to disable mouse input. I've found it to be a handy configuration.


mahoneyt944:
I'll text later and post back.

mahoneyt944:
Ok first glance.

- missing parts of player1, button 4 defaults
- I wouldn't map the gun to IPT_JOYSTICKLEFT_* because this is for dual joystick games and wouldn't be logical.
- I think we only support 6 players right now so the 7 and 8 player additions probably aren't needed.

markwkidd:
Thank you for catching the P1 BUTTON4 issue! What you are saying about not binding the Left/Left joystick to the gun makes sense. I've pushed commits for both of those.

Before I go ahead and remove the Player 7 and 8 gun bindings, would it make sense for me to go ahead and add default retropad and mouse for P7 and P8 instead? I think there is at least one game supported by the core that goes up to 7 players but maybe 6 is where it tops out.

mahoneyt944:
I'm not sure if we support any 7 and 8 player games. Do you know of any? Lol. I know we boosted to support 6 players for the X-Men games.

markwkidd:
I guess it's sprint8 that we have, with 8 players. Hm.

markwkidd:
Updates

* **3 default gun mappings.** It doesn't seem like there are any games with more than three lightguns, so in the latest commit there are only default gun mappings for the first three players.
* **P7 & P8** I added support for players 7 and 8, along with default mouse mappings up to 8, because Sprint 8 has mouse-like input for that many players. I haven't tested that game.

PR #1057, opened 2021-05-25
remove unused ctrlr string parsing code

markwkidd:
This code is unused in our core. I looked at MAME 0.78 and it is involved in parsing ctrlr files. As far as I can tell, we don't have any potential future use for this, and it will diverge from our active code unless I add strings to the struct for the new lightgun codes. 

Therefore I think we could just get rid of this code. What do you all think?


mahoneyt944:
I've never looked into this part of the code. Maybe grant has at some point? Might be a person to verify with. Other than that heavily test first? Ha

grant2258:
yes this is for parsing controller assignments. It can be removed personally I had kept it updated for a reason one day someone might use it to set the retropad in the local machine. The real problem you have atm is your using input general instead of the machine mapping which can be different if the driver doesnt set the default or the user changes it. I personally removed the input macros on my port its not very helpful if you need to manipulate things manually. All roads lead to rome though.

mahoneyt944:
@grant2258 would it be easier to assign these games more typically mappings or are these odd ball games that have unusual inputs?

mahoneyt944:
I don't really have a preference on this code change. As long as it doesn't effect the current code base negatively.

markwkidd:
Grant! Good to see you, it's been a while.

As you pointed out, the mappings I've been adding so far are generic ones, mostly oriented towards lightguns. That's not in this PR though. Nothing I have done is content-aware, which would be better :D The code you linked does look like the smart way to handle those machines, thank you

In terms of this PR, we seem unanimous that the code affected was used in parsing controller files, so I will proceed to merge it

PR #1060, opened 2021-05-28
Lightgun option

mahoneyt944:
changes "Mouse Device" core option label to "X-Y Device".
polls either mouse, pointer, lightgun or none by core option.


markwkidd:
Thank you. Maybe there is a way to make this more seamless so that a core option isn't needed in the future, but that's a small detail compared to having the basic lightgun functionality working IMO.

PR #1064, opened 2021-05-28
7 and 8 players

mahoneyt944:
Adding more stuff for 7 and 8 players. Do we have enough ports for 8 players??


markwkidd:
I can't check sprint8 right now for P7 and P8 but that will be on my TODO list

StormedBubbles:
sprint8 works with 8 players! Thank you for adding this! Each player has his/her own coin slot, "RetroPad" A = brake, "RetroPad" B = accelerate, and the space bar = change track. Each player is also supposed to have a gear-shift mechanism (low/high), but I'm not sure if that is included. From looking at the controls, it may be the x-axis on the right analog stick, but it is hard to tell.

mahoneyt944:
Great. Thank you for testing

PR #1070, opened 2021-06-12
temp descriptor patch

grant2258:
This should fix you up for now. Just change as needed and close this


mahoneyt944:
I don't think we have any lightgun as device parent anymore. Driving so can't look now.

mahoneyt944:
thank you

PR #1074, opened 2021-06-20
clean up makefile, revert android mouse/gun input workaround

markwkidd:
Why do I invite the wrath of the gods by tampering with a makefile?

Well, some of the comments are plain wrong, and compiler flags which get applied to many systems are added in at various times. I'm in no hurry to merge this and I'll definitely be keeping an eye on the automated build :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -709,6 +709,13 @@ else
 endif
 
 
+# Architecture-specific flags #############################
+ifeq ($(BIGENDIAN), 1)
+	PLATCFLAGS += -DMSB_FIRST
+endif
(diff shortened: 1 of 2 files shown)


mahoneyt944:
@markwkidd at a glance it looks good to me. This PR seems to include the changes proposed in #1075 so we could probably close that one and merge this one. As far as the android work around I think it makes sense to push that workaround to the front end that way it can be fixed properly.

markwkidd:
Well dang, I missed a git reset --hard
I guess it is all in one place. I'll give this PR a better name for the log and merge. Thanks.

PR #1081, opened 2021-06-21
pass RETRO_PROFILE from Makefile to frontend

markwkidd:
All the other cores pass this value with by reference, so I figured let's explicitly cast to unsigned, and give it a variable instead of a literal. I want to make sure this compiles everywhere before declaring victory.


mahoneyt944:
Tested working on Android now.

PR #1098, opened 2021-07-09
Added Core option to disable cyclone

apison:
Added Core option to disable cyclone

### DIFF
--- a/src/mame.h
+++ b/src/mame.h
@@ -252,7 +252,7 @@ struct GameOptions
   bool     machine_timing;
   bool     digital_joy_centering; /* center digital joysticks enable/disable */
 #if (HAS_CYCLONE || HAS_DRZ80)
-  int      enable_cyclone;
+  int      cyclone_mode;
 #endif
 };
 
--- a/src/mame.h
+++ b/src/mame.h
@@ -251,9 +251,9 @@ struct GameOptions
   bool     cheat_input_ports;     /*cheat input ports enable/disable */
   bool     machine_timing;
   bool     digital_joy_centering; /* center digital joysticks enable/disable */
-  #if (HAS_CYCLONE || HAS_DRZ80)
+#if (HAS_CYCLONE || HAS_DRZ80)
   int      enable_cyclone;
-  #endif
+#endif
   };
 
 
--- a/src/mame.h
+++ b/src/mame.h
@@ -251,6 +251,9 @@ struct GameOptions
   bool     cheat_input_ports;     /*cheat input ports enable/disable */
   bool     machine_timing;
   bool     digital_joy_centering; /* center digital joysticks enable/disable */
+  #if (HAS_CYCLONE || HAS_DRZ80)  
+  int	   enable_cyclone;	
+  #endif					
   };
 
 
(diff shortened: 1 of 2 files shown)


mahoneyt944:
@grant2258 what do you think about moving init_core_options() in this PR?

Also seems repetitive to set options.enable_cyclone in retro_load_game. Is this necessary or should we just add the (restart core) like the other core options?

Thoughts?

grant2258:
Im guessing only the op didnt want to be invasive with the code he was adding. This code can be re factored to be added here

https://github.com/libretro/mame2003-plus-libretro/blob/00f5bab589865ea3f5cd9a8596a6b2ac6bc39f73/src/mame2003/mame2003.c#L287

https://github.com/libretro/mame2003-plus-libretro/blob/00f5bab589865ea3f5cd9a8596a6b2ac6bc39f73/src/mame2003/mame2003.c#L416

mahoneyt944:
And it appears that this is uneeded. As it should get cause in the else's default below. Correct?
```
  /* none - cyclone core option */
  else if (options.enable_cyclone==0)
  {
    use_cyclone = 0;
    use_drz80_snd = 0;
    use_drz80 = 0;
  }
```

grant2258:
nah it is needed because its all set 1 as default and disables as needed.

apison:
Im not a c++ programmer and above all im not a retroarch expert so all the modify that improve the option are very well accepted.

mahoneyt944:
@grant2258 if we removed that section and the core option was set to "none" (0). That means the "cores" variable would = -1 so no case would match and the default catch sets the same thing? Or am I seeing things wrong?
```
      default:
        use_cyclone = 0;
        use_drz80_snd = 0;
        use_drz80 = 0;
```

```
  /* none - cyclone core option */
  else if (options.enable_cyclone==0)
  {
    use_cyclone = 0;
    use_drz80_snd = 0;
    use_drz80 = 0;
  }

  /* Force cyclone with user selected core option */
  else
  {
    int cores=0;
    cores=options.enable_cyclone-1;

    switch (cores)
    {
      case 0:
        use_cyclone = 0;
        use_drz80_snd = 0;
        use_drz80 = 0;
        break;
      case 1:
        use_drz80_snd = 0;
        use_drz80 = 0;
        break;
      case 2:
        use_cyclone = 0;
        break;
      case 4:
        use_cyclone = 0;
        use_drz80 = 0;
        break;
      case 5:
        use_drz80 = 0;
        break;
      default:
        use_cyclone = 0;
        use_drz80_snd = 0;
        use_drz80 = 0;
        break;
    }
  }
```

grant2258:
code looks good its wrapped round an if else to catch it. It does need re factored though. Problem is this only changes on reloading the core instead of real time.

apison:
The default settings should be used to use the  fe_drivers struct cyclone configuration that in most cases is the correct ones.

mahoneyt944:
@apison I updated the default core option already to use the default first

grant2258:
I dont know really I personally think this makes the list redundant anyway. Do we really need all them options on or off for each core should be fine. You would need to tie in with mark on this. From what i can see @apison  added the feature he wanted and still lets the list work if that is wanted. I think this is confuing for users though on and off is more clear.

mahoneyt944:
I think the idea is to be able to override the default cyclone settings and experiment with what works? It's disabled for most platforms anyhow. 

I think it just needs really optimized / cleaned up.

mahoneyt944:
This PR should function the same as the core does now except the user can override the default cyclone settings using the core option now. The limitation right now is the core needs a full reboot after changing this setting for it to take effect.

PR #1102, opened 2021-07-17
fix old frameskip and add new

grant2258:
@mahoneyt944 please test this before adding should be fine just make sure


mahoneyt944:
Looks good thanks.

grant2258:
Np

PR #1108, opened 2021-07-29
fix #1053 save hiscore periodically

### LINKED ISSUE #1053: Save hiscore files regularly to prevent loss of hiscore when exiting abnormally
At present, hiscore is only saved when mame exits or resets. On embedded platforms, the power of the machine is usually turned off directly, which will cause hiscore to be lost.

LuoZhongYao:
Save hiscore every 1000 frames to prevent hiscore from being lost when exiting abnormally

### DIFF
--- a/src/hiscore.c
+++ b/src/hiscore.c
@@ -10,12 +10,15 @@
 #include "../precompile/hiscore_dat.h"
 
 #define MAX_CONFIG_LINE_SIZE 48
+#define HISCORE_SYNC_FILE_DELAY	1000
 
 const char *db_filename = "hiscore.dat"; /* high score definition file */
 
 static struct
 {
 	int hiscores_have_been_loaded;
+	int hs_sync_delay;
+	mame_file *hs_file;
 
 	struct mem_range
 	{
@@ -162,6 +165,7 @@ static void hs_load (void)
 {
 	mame_file *f = mame_fopen (Machine->gamedrv->name, 0, FILETYPE_HIGHSCORE, 0);
 	state.hiscores_have_been_loaded = 1;
+	state.hs_sync_delay = HISCORE_SYNC_FILE_DELAY;
     
 	if (f)
 	{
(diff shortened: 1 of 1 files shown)


mahoneyt944:
I'm a little reluctant to merge this because it effects every game using high scores and that would be an excessive amount of testing to do for anyone. The code looks ok to me though. I do wonder if this will cause any performance hits or possibly corrupted saves.... Some saves require user actions such as starting a new game to update the field properly.

Being that this is only one file being changed, it's easily to revert if we see any regressions so I'll merge it on the premise that we'll need some user  feed back.

grant2258:
This should be optional doing this every 16.6 seconds seems excessive for people that actatually exit anyway. It's really needless unless people want it others might want an option  not to kill there hard drives because someone else turns the box off and can't be bothered exiting .

mahoneyt944:
Could add a core option? Auto save  highscore

mahoneyt944:
I think we could just add a option check here
https://github.com/libretro/mame2003-plus-libretro/blob/2e144d99cb57a0a4b5e5c2d715d342b457f29d89/src/hiscore.c#L339

grant2258:
Is really up to you personally any exiting should be working anyway or fixed if there is an issue with a specific game. This is bad practice you should name it recursively save hi scores so users know its constantly saving. That way no one can complain if its on.

mahoneyt944:
The issue was that the user is killing power to the device so no exiting code is ever called to save so this is a workaround. Agreed not ideal. Check the closed issue on this for more details, maybe you see another solution?

mahoneyt944:
I was thinking we could have option:
save hiscore ....on exit, every 10 sec, every 30 sec, every 1 min etc

mahoneyt944:
@grant2258 do you think this is clear enough?
https://github.com/libretro/mame2003-plus-libretro/compare/Optional

grant2258:
Well as long as its optional it's all good. I have my raspberry pi and it has no on off issue as well but I always shut it down before powering off. The same thing would happen with a pc if they used the reset button. I think this user behaviour rather rather than anything else. You run the risk of corrupting data just unplugging on any device. The ones lacking a power button should still be shutdown via other methods.

LuoZhongYao:
I think you can set autosave_hiscore to a value in units of frames, change HISCORE_SYNC_FILE_DELAY to autosave_hiscore, turn off autosave hiscore, you can set autosave_hiscore to INT_MAX or LONG_MAX, so you can adjust the save interval, the disadvantage is that the options are not intuitive

mahoneyt944:
I merged a core option. I think it's clear enough default / recursively. Needs tested though.

PR #1115, opened 2021-08-08
Isolation of core options in prep for new API

mahoneyt944:
@markwkidd @grant2258 
could you guys review this PR? I figured we should separate the core options before unloading the new v2 core option structure on it. This should function exactly the same as it does now. Both of you guys have done lots of work in this area so I figured one of you maybe able to detect any potential issues. 🙃


mahoneyt944:
@markwkidd @grant2258 
could you guys review this PR? I figured we should separate the core options before unloading the new v2 core option structure on it. This should function exactly the same as it does now. Both of you guys have done lots of work in this area so I figured one of you maybe able to detect any potential issues. 🙃

PR #1149, opened 2021-09-22
Helifire

mahoneyt944:
VIDEO_EOF( helifire ) does not seem to set the palette colors correctly. Only i = 0 is being set....this is why all the object colors are black as far as I can tell. Can't seem to figure out why this is the case though. The code seems correct so I believe something else is interfering with the palette at these values (0 - 7) or we are simply missing something somewhere.

mame2003
https://github.com/libretro/mame2003-libretro/blob/97bc249ad679258bcf5e8a17e86b8626a9f89b18/src/vidhrdw/8080bw_vidhrdw.c#L639-L693

mame84
https://github.com/mamedev/historic-mame/blob/683b199d73bf3f692d7fddf24ab6d45ba51544bf/src/vidhrdw/n8080.c#L447-L482

Also some audio changes from
https://github.com/mamedev/mame/commit/5cd4c55b9b94d7e59c325ef6a0e1ba1d6583228d


mahoneyt944:
@grant2258 @arcadez2003 
either of you see anything out of place in regards to the palette colors set at 0 through 7? Or why they don't seem to be working?

mahoneyt944:
@arcade Nevermind I got it. For some reason it likes the old rom loading with the new code.

PR #1168, opened 2021-10-16
pointer device should be analog do not pull yet

grant2258:
Need to make some udev updates in retroarch to let the mouse return a valid pointer pressed when its not a touch device. We need to start doing things the way the api does expects it

### DIFF
--- a/src/inptport.c
+++ b/src/inptport.c
@@ -1491,11 +1491,11 @@ profiler_mark(PROFILER_INPUT);
 		osd_analogjoy_read (i, analog_current_axis[i], analogjoy_input[i]);
 
 		/* update mouse/trackball position */
-		if(options.xy_device == RETRO_DEVICE_MOUSE || options.xy_device == RETRO_DEVICE_POINTER)
+		if(options.xy_device == RETRO_DEVICE_MOUSE )
 			osd_xy_device_read (i, &(mouse_delta_axis[i])[X_AXIS], &(mouse_delta_axis[i])[Y_AXIS]);
 
 		/* update lightgun position, if any */
-		else if(options.xy_device == RETRO_DEVICE_LIGHTGUN)
+		else if(options.xy_device == RETRO_DEVICE_LIGHTGUN || options.xy_device == RETRO_DEVICE_POINTER )
  			osd_xy_device_read (i, &(lightgun_delta_axis[i])[X_AXIS], &(lightgun_delta_axis[i])[Y_AXIS]);
 	}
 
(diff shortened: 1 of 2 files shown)


mahoneyt944:
I guess in this case we can remove get_pointer_delta completely correct?

mahoneyt944:
How many platforms use an actual pointer? Idk if I have one or not

grant2258:
The pointer device read is a Retroarch thing. All mice and touchpads are pointers some are relative and some are absolute both should be supported at driver level and return the appropriate output.

Some touchpads on laptops, all touchscreens are abs devices it would depend on the device touchpads are harder answer because they could be rel or abs. 

This shouldnt matter to the core the driver should convert and send the appropriate settings. Thats exactly what my pr does for Ra that check is the mouse abs or relative return the button type.

mahoneyt944:
RA commit merged, is this ready then?

grant2258:
compiling myself its really odd this one hoing to try on another pc

grant2258:
I even compiled ra 1.9.10 to rule that out the mouse is working on my core. Ill compile on another pc i havent used this core in a while what os are you testing the xy devices on im not even seeing the cross hairs

mahoneyt944:
I'm using Android v8a and a rpi3 build

grant2258:
There is defo something funky going on here. I matched my other linux box commits that work. With this repo and and the same with RA and its the same my core works plus doesnt so the mouse is and pointer is defo working. Looks like something is really amiss here somewhere.

mahoneyt944:
This is odd,
The joycode should read 1000s for player1 and 2000s for player2 and so on?
Screenshot_20211016-223205

grant2258:
not getting any of that logging at all you sure your on the latest update

mahoneyt944:
You have to log with debug level

mahoneyt944:
That's what I do, log to file. Just start a game and close it right away or else it will be super long. Then search for joycode.

mahoneyt944:
That's an old build. Unsupported controllers aren't in the build anymore.

grant2258:
joycode has nothing to do this the mouse not working though

grant2258:
You said its working for you though so it must be something else

mahoneyt944:
The loop has seemed odd ever since it was changed though. And now that I removed the unsupported controllers it's showing more noticably in the debug log. Just need to see what's happening. Not sure what's going on with mouse stuff that seems to be working for me. This is all very odd.

grant2258:
The core probes that when it starts I tried another core with mouse support and thats working fine too. Its looking like it might be this core. I guess ill need to set the pi up and check there

PR #1169, opened 2021-10-17
fix the mouse recalibrating in operation wolf

grant2258:
@mahoneyt944 this should fix us up for the issue we had before. With othunder the relative and abs have been seperated are still on one function though.

points to note
a mouse/pointer device can be rel or abs this shouldnt matter to the core. The driver should send the appropriate data back when we request if no need for us to do any conversion.

input_cb(player, RETRO_DEVICE_POINTER, 0, RETRO_DEVICE_ID_POINTER_X)
input_cb(player, RETRO_DEVICE_MOUSE, 0, RETRO_DEVICE_ID_MOUSE_X)
input_cb(player, RETRO_DEVICE_LIGHTGUN, 0, RETRO_DEVICE_ID_LIGHTGUN_SCREEN_X)

we just read what data we need from that device its down to the frontend to handle it properly and send it back if.

@markwkidd  @mahoneyt944  if there is any issues on platforms ive tested on udev let me know ill look at it then we can post an issue on RA to get the drivers in shape or get help doing it so it all works like it should.

### DIFF
--- a/src/inptport.c
+++ b/src/inptport.c
@@ -1491,10 +1491,10 @@ profiler_mark(PROFILER_INPUT);
 		osd_analogjoy_read (i, analog_current_axis[i], analogjoy_input[i]);
 
 		/* update mouse/trackball position */
-		osd_xy_device_read (i, &(mouse_delta_axis[i])[X_AXIS], &(mouse_delta_axis[i])[Y_AXIS],0);
+		osd_xy_device_read (i, &(mouse_delta_axis[i])[X_AXIS], &(mouse_delta_axis[i])[Y_AXIS], "relative");
 
 		/* update lightgun position, if any */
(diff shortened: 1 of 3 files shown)


grant2258:
There is an issue when switching between abs and back to rel in the mame code it doesnt effect the games calibration but the sensitivity goes up. Mames code base was never meant to do this though. so it would be best to always just read lightgun games as a pointer or lightgun I would say default to lightgun. I can see any use for the pointer device to be honest since we have the lighgun will leave it to you guys what you want to do with you must always read the mouse. The code causing the issue is here id you want to update it for a graceful transition.

grant2258:
I can do much about the nv ram needed deleted but we wont get this situation again thankfully. It working and tested as is if you wanna change it anything go for it.

grant2258:
looks fine to me need to set a pie up and rest some stuff now.

PR #1170, opened 2021-10-17
Isolate relative and absolute calls

mahoneyt944:
This should allow us to control behaviors more finely when absolute devices are selected while relative calls are made and vice versa. This should also help to ensure the structures have some predictable default behavior so we can properly override controls and apply scaling and such. @grant2258 we'll have to figure out how we can rework the scaling to be active with our device selection and this should be buttoned up

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -215,7 +215,6 @@ android-arm64-v8a:
     - .core-defs
   only:
     - master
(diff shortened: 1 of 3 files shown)


grant2258:
There Really is no need for this they are already isolated  your just disabling mouse/trackballs/dials if you dont select device mouse

PR #1197, opened 2021-11-21
use libretro rotations when available and let mame sort anything that need roated

grant2258:
This code simply does rotations when need a user issue inspired me to look into this again. Only tested with disabled rotation in settings->core works as intended you will need RA 1.9.13.2 at least for the commit to be in. Im not sure how the wii port is working if it will return a fail or not when teh rotation doesnt happen if it doesnt just force it to not allow rotation in the setting mentioned above.

### DIFF
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -142,7 +142,6 @@ void mame2003_video_init_orientation(void)
       video_hw_transpose = orientation & ORIENTATION_SWAP_XY;
       orientation = 0;
    }
-
    /* Otherwise try to use it to do a transpose */
    rotate_mode = 3; /* ROT90 */
    if (orientation & ORIENTATION_SWAP_XY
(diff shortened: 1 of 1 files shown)


mahoneyt944:
Should we confirm a test with the wiiu before merging? Or any other platforms troubled by this?

grant2258:
Sure if you like you can initiate it by disabling rotation any any platform

mahoneyt944:
Ummm I think big endian? Not sure.

Ploggy:
> I think big endian.

Yes, the WiiU is Big Endian :) 
Can you Merge this so I can test the Buildbot version? I'm getting crashes every time I load 2 Games (a Mame2003Plus Game, then any other Game from any Core) when I compile Mame2003plus myself.

I dont mind waiting till the fix build hits the buildbot.

mahoneyt944:
Surely

mahoneyt944:
The crash issue may be just be a memory issue with the platform. I've heard about this with 3ds once before too. Can you load a rom from another core then load  a mame2003-plus rom after?

mahoneyt944:
@grant2258 @Ploggy  this seems to break a lot of rotations on my Android build. 1941, 1943, food fight, jail break to name a few. I'm also using RA 1.9.13 stable I believe. Not the nightly.

Ploggy:
> The crash issue may be just be a memory issue with the platform. I've heard about this with 3ds once before too. Can you load a rom from another core then load a mame2003-plus rom after?

Loading any Core/Game then Mame2003 Plus is fine, but as soon as I load any Game after a Mame2003Plus game the Console crashes...
Though only with my builds it seems, latest nightly Mame2003 Plus doesn't crash when taking the same steps..

If the fix is not ready a good quick fix for me is to just disable rotation Settings>Core, it doesn't completely fix the issue but it does stop the Ozone UI from rotating :)  more specifically the UI doesn't rotate the UI selection box does....

https://drive.google.com/file/d/1LFLnsmP7npSxzo87xG8wN1gcaKX22qg_/view?usp=sharing

grant2258:
@Ploggy the Retorarch rotation doesnt work on on your port. for this patch to work you simply  Allow Rotation "off" in Settings>Core and settings->video->output should be normal. The your vertical games should show up properly with this patch with no other interaction. 

normal mame 2003 will will show the old behavior and the aspect ratio will be wrong which is normal since we used retroarch for the rotation

grant2258:
@Ploggy I can confirm from your video your mame 2003-plus dint have my commit

PR #1199, opened 2021-11-23
Wii u rotation fix

grant2258:
@mahoneyt944 add this till the user get the build he needs if you dont mind and let him confirm he has it before reverting if you are going to revert it. I have no idea what going on the android end it shouldnt be causing issues though. That


mahoneyt944:
This should be ok for now.

grant2258:
There is some issues with the video code conditons will sort this at the weekend am on backshift. It there any way to just put these in a test branch?

PR #1202, opened 2021-11-28
fix rotations when rotations are disabled or not supported

grant2258:
Had time to fix this up properly at teh weekend. So RA will do the rotations if available. Its up to the frontend to report the rotations are unsupported and the recent addition to RA sets the disable option like it should. If the frontend doesnt support rotations failures there is no change from previously it.

### DIFF
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -126,7 +126,7 @@ void mame2003_video_init_orientation(void)
    {
       if (environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode) )
       {
-        log_cb(RETRO_LOG_INFO,"retroarch doing the rotation\n");
+        log_cb(RETRO_LOG_INFO, LOGPRE "RetroArch will perform the rotation\n");
 
         rotate_mode = (orientation == ROT270) ? 1 : rotate_mode;
         rotate_mode = (orientation == ROT180) ? 2 : rotate_mode;
@@ -137,11 +137,11 @@ void mame2003_video_init_orientation(void)
       }
 
       else
-        log_cb(RETRO_LOG_INFO,"This port of Retroarch does not support rotation or its been disabled mame will rotate internally\n");
+        log_cb(RETRO_LOG_INFO, LOGPRE "This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally\n");
 
    }
    else
-     log_cb(RETRO_LOG_INFO,"retroarch does not support this type of rotation using mame internal rotation\n");
+     log_cb(RETRO_LOG_INFO, LOGPRE "RetroArch does not support this type of rotation, using mame internal rotation instead\n");
 
    tate_mode = options.tate_mode;
 
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -62,6 +62,7 @@ void mame2003_video_get_geometry(struct retro_game_geometry *geom)
    geom->base_height = video_hw_transpose ? vis_w : vis_h;
 
    geom->aspect_ratio = video_hw_transpose ? (float)video_config.aspect_y / (float)video_config.aspect_x : (float)video_config.aspect_x / (float)video_config.aspect_y;
+
 }
 
 void mame2003_video_update_visible_area(struct mame_display *display)
@@ -110,63 +111,49 @@ void mame2003_video_init_orientation(void)
    unsigned orientation = Machine->gamedrv->flags & ORIENTATION_MASK;
    unsigned rotate_mode;
 
-   /* Acknowledge that the TATE mode is handled */
-   tate_mode = options.tate_mode;
+   rotate_mode = 0; /* Known invalid value */
+   tate_mode = options.tate_mode;   /* Acknowledge that the TATE mode is handled */
+   video_hw_transpose = false;
 
-   /*
-      The UI is always oriented properly at start, but the externally applied
-      orientation needs to be countered.
-   */
+   /* test RA if rotation is working if not dont alter the orientation let mame handle it */
    options.ui_orientation = reverse_orientation(orientation);
 
-   /* Do a 90 degree CCW rotation for vertical games in TATE mode */
    if (tate_mode && (orientation & ORIENTATION_SWAP_XY))
       orientation = reverse_orientation(orientation) ^ ROT270;
 
-   /* Try to reset libretro orientation */
-   rotate_mode = 0;
-   environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode);
 
-   /* Try to match orientation to a supported libretro rotation */
-   rotate_mode = 0; /* Known invalid value */
-   rotate_mode = (orientation == ROT270) ? 1 : rotate_mode;
-   rotate_mode = (orientation == ROT180) ? 2 : rotate_mode;
-   rotate_mode = (orientation == ROT90) ? 3 : rotate_mode;
-
-   video_hw_transpose = false;
-
-   /* Try to use libretro to do a rotation */
-   if (rotate_mode != 0 ) /* Known invalid value */
-      if (environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode))
-      {
-         video_hw_transpose = orientation & ORIENTATION_SWAP_XY;
-         orientation = 0;
-      }
-   /* Otherwise try to use it to do a transpose */
-   if (rotate_mode == 3 && orientation & ORIENTATION_SWAP_XY) /* ROT90 */
-      if (environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode))
-      {
-         video_hw_transpose = true;
-         orientation = reverse_orientation(orientation ^ ROT270);
-      }
-   #ifdef WIIU
-  /*If Ra fails to rotate through settings->core  or doesnt implement rotation adjust for it recent addition to the code allows this check*/
-   if (!environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode))
+   if (orientation == ROT0 || orientation == ROT90 || orientation == ROT180 || orientation == ROT270)
    {
-      
-      log_cb(RETRO_LOG_INFO,"RA rotation failed mame will assume its duties\n"); 
+      if (environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode) )
+      {
+        log_cb(RETRO_LOG_INFO,"retroarch doing the rotation\n");
 
-      if (orientation & ORIENTATION_SWAP_XY )   
-         video_hw_transpose = true;
+        rotate_mode = (orientation == ROT270) ? 1 : rotate_mode;
+        rotate_mode = (orientation == ROT180) ? 2 : rotate_mode;
+        rotate_mode = (orientation == ROT90) ? 3 : rotate_mode;
+        if (orientation & ORIENTATION_SWAP_XY) video_hw_transpose = true; /*do this before the rotation reverse*/
+        orientation = reverse_orientation(orientation ^ orientation); /* undo mame rotation if retroarch can do it */
+        environ_cb(RETRO_ENVIRONMENT_SET_ROTATION, &rotate_mode);
+      }
 
-    }
-    #endif
+      else
+        log_cb(RETRO_LOG_INFO,"This port of Retroarch does not support rotation or its been disabled mame will rotate internally\n");
+
+   }
+   else
+     log_cb(RETRO_LOG_INFO,"retroarch does not support this type of rotation using mame internal rotation\n");
+
+   tate_mode = options.tate_mode;
+
    /* Set up native orientation flags that aren't handled by libretro */
+   if (orientation & ORIENTATION_SWAP_XY) video_hw_transpose = true; /*dont set this to false if RA changes the flags else vertical games swap the xy*/
    video_flip_x = orientation & ORIENTATION_FLIP_X;
    video_flip_y = orientation & ORIENTATION_FLIP_Y;
    video_swap_xy = orientation & ORIENTATION_SWAP_XY;
-
+   log_cb(RETRO_LOG_INFO,"mame internal: video_flip_x:%u video_flip_y:%u video_swap_xy:%u video_hw_transpose:%u\n",video_flip_x,video_flip_y,video_swap_xy,video_hw_transpose);
    Machine->ui_orientation = options.ui_orientation;
+
+
 }
 
 /* Init video format conversion settings */
@@ -236,7 +223,7 @@ void mame2003_video_reinit(void)
    UINT32 rgb_components[3];
    struct osd_create_params old_params = video_config;
    osd_close_display();
-   osd_create_display(&old_params, &rgb_components);
+   osd_create_display(&old_params, &rgb_components[0]);
 }
 
 int osd_create_display(
@@ -251,7 +238,6 @@ int osd_create_display(
    video_do_bypass =
       !video_flip_x && !video_flip_y && !video_swap_xy &&
       ((video_config.depth == 15) || (video_config.depth == 32));
-
    /* Allocate an output video buffer, if necessary */
    if (!video_do_bypass)
    {


vaguerant:
Testing this out on Wii U with Allow Rotation set to off because rotation is hooked up but has broken output on Wii U. It seems like the core is still reporting the unrotated aspect ratio to RetroArch, e.g. using Core Provided aspect ratio in Settings > Video > Scaling is still giving 4:3 for games which are actually 3:4, like Donkey Kong. Am I doing something wrong here, do I need to configure something else to get MAME 2003 Plus to report the post-rotation aspect ratio?

grant2258:
youll need https://github.com/libretro/RetroArch/commit/211e7d4094bdf16af5e86db74590627c05b3835a  for this to work properly.

vaguerant:
Aha, I've just realized it's not Core Provided which has issues, but Custom. Core Provided is giving the correct aspect ratio, but Custom is still passing RetroArch the original un-rotated resolution. e.g. If I set Arkanoid to 1x integer scale in MAME 2003 Plus, it reports as Width: 256 and Height: 224, when it should be the opposite.

Here's a log for Arkanoid:
```
[INFO] RetroArch 1.9.13 (Git fbbe351)
[INFO] === Build =======================================
[INFO] Capabilities: 
[INFO] Built: Dec  4 2021
[INFO] Version: 1.9.13
[INFO] Git: fbbe351
[INFO] =================================================
[INFO] [Input]: Found input driver: "wiiu".
[INFO] [Overrides]: No core-specific overrides found at "sd:/retroarch/config/MAME 2003-Plus/MAME 2003-Plus.cfg".
[INFO] [Overrides]: No content-dir-specific overrides found at "sd:/retroarch/config/MAME 2003-Plus/Rotation.cfg".
[INFO] [Overrides]: No game-specific overrides found at "sd:/retroarch/config/MAME 2003-Plus/arkanoid.cfg".
[INFO] [Remaps]: Remap directory: "sd:/retroarch/config/remaps".
[INFO] [Overrides]: Redirecting save file to "sd:/retroarch/cores/savefiles/MAME 2003-Plus/arkanoid.srm".
[INFO] [Overrides]: Redirecting save state to "sd:/retroarch/cores/savestates/MAME 2003-Plus/arkanoid.state".
[INFO] [Environ]: GET_LOG_INTERFACE.
[INFO] [Environ]: PERFORMANCE_LEVEL: 10.
[INFO] [Content]: Content loading skipped. Implementation will load it on its own.
[libretro INFO] [MAME 2003+] Full content path sd:/Game Backups/Arcade/MAME 2003-Plus/Rotation/arkanoid.zip
[libretro INFO] [MAME 2003+] Git Version  379c0e9
[libretro INFO] [MAME 2003+] Content lookup name: arkanoid
[libretro INFO] [MAME 2003+] Driver index counter: 1271. Matched game driver: arkanoid
[libretro INFO] [MAME 2003+] ==== BEGIN DRIVER CONTENT ATTRIBUTES ====
[libretro INFO] [MAME 2003+] * Supports 2 players with 2 distinct controls.
[libretro INFO] [MAME 2003+] * Supports 1 distinct button controls.
[libretro INFO] [MAME 2003+] * Uses a dial.
[libretro INFO] [MAME 2003+] * Uses a tilt function.
[libretro INFO] [MAME 2003+] * Uses multiplayer control labels.
[libretro INFO] [MAME 2003+] * Uses 8-way joystick controls.
[libretro INFO] [MAME 2003+] ==== END DRIVER CONTENT ATTRIBUTES ====
[INFO] [Environ]: SYSTEM_DIRECTORY: "sd:/retroarch/cores/system".
[INFO] [Environ]: GET_SAVE_DIRECTORY.
[libretro INFO] [MAME 2003+] content path: sd:/Game Backups/Arcade/MAME 2003-Plus/Rotation
[libretro INFO] [MAME 2003+]  system path: sd:/retroarch/cores/system
[libretro INFO] [MAME 2003+]    save path: sd:/retroarch/cores/savefiles/MAME 2003-Plus
[INFO] [Environ]: GET_CORE_OPTIONS_VERSION.
[INFO] [Environ]: RETRO_ENVIRONMENT_SET_CORE_OPTIONS_V2.
[INFO] [Environ]: SET_AUDIO_BUFFER_STATUS_CALLBACK.
[INFO] [Environ]: GET_LED_INTERFACE.
[INFO] [Environ]: SET_CONTROLLER_INFO.
[libretro INFO] [MAME 2003+] Opening ROM file: a75_01-1.rom
[libretro INFO] [MAME 2003+] Opening ROM file: a75_11.rom
[libretro INFO] [MAME 2003+] Opening ROM file: arkanoid.uc
[libretro INFO] [MAME 2003+] Opening ROM file: a75_03.rom
[libretro INFO] [MAME 2003+] Opening ROM file: a75_04.rom
[libretro INFO] [MAME 2003+] Opening ROM file: a75_05.rom
[libretro INFO] [MAME 2003+] Opening ROM file: 07.bpr
[libretro INFO] [MAME 2003+] Opening ROM file: 08.bpr
[libretro INFO] [MAME 2003+] Opening ROM file: 09.bpr
[libretro INFO] [MAME 2003+] Succesfully loaded ROMs.
[INFO] [Environ]: SET_ROTATION: 0
[libretro INFO] [MAME 2003+] This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally
[libretro INFO] mame internal: video_flip_x:1 video_flip_y:0 video_swap_xy:1 video_hw_transpose:1
[INFO] [Environ]: SET_PIXEL_FORMAT: RGB565.
[libretro INFO] [MAME 2003+] 

CONTROLS: 

GAMEINFO: Arkanoid (World)
1986 Taito Corporation Japan

CPU:
Z80 6.000000 MHz
M68705 500.000 kHz

Sound:
1xAY-3-8910 1.500000 MHz

Screen Resolution:
256 x 224 (V) 60.000000 Hz

[INFO] [SRAM]: Skipping SRAM load.
[INFO] [Core]: Version of libretro API: 1, Compiled against API: 1
[libretro INFO] [MAME 2003+] Sample rate set to 48000
[INFO] [Audio]: Set audio input rate to: 47952.05 Hz.
[INFO] [Video]: Set video size to: 879x768.
[INFO] [hid]: initializing
[INFO] [hid]: starting polling thread.
[INFO]                 handle: 1924281
[INFO]   physical_device_inst: 0
[INFO]                    vid: 0x6d04
[INFO]                    pid: 0x2bc5
[INFO]        interface_index: 2
[INFO]              sub_class: 0
[INFO] [INFO]              protocol: 0
    max_packet_size_rx: 32
[INFO]     max_packet_size_tx: 0
g
[INFO] [INFO]                 handle: 1970794
[INFO]   physical_device_inst: 0
[INFO]                    vid: 0x6d04
[INFO]                    pid: 0x2bc5
[INFO]        interface_index: 1
[INFO]              sub_class: 1
[INFO]               protocol: 2
[INFO]     max_packet_size_rx: 8
[INFO]     max_packet_size_tx: 0
[INFO]                 handle: 2038587
[INFO]   physical_device_inst: 0
[INFO]                    vid: 0x6d04
[INFO]                    pid: 0x2bc5
[INFO]        interface_index: 0
[INFO]              sub_class: 1
[INFO]               protocol: 1
[INFO]     max_packet_size_rx: 8
[INFO]     max_packet_size_tx: 0
[INFO] [Input]: Found HID driver: "wiiu".
[INFO] [Joypad]: Found joypad driver: "wiiu".
[INFO] [Video]: Setting refresh rate to: 59.940 Hz.
[INFO] [Audio]: Set audio input rate to: 47952.05 Hz.
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Video]: Found display server: "null".
[INFO] [Display]: Found display driver: "gx2".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Font]: Using font rendering backend: "stb-unicode".
[INFO] [Environ]: SET_INPUT_DESCRIPTORS:
[INFO] [Playlist]: Loading history file: "sd:/retroarch/content_history.lpl".
[INFO] [Playlist]: Loading history file: "sd:/retroarch/content_music_history.lpl".
[INFO] [Playlist]: Loading favorites file: "sd:/retroarch/content_favorites.lpl".
[INFO] Failed to find entry for vid: 0x6d04, pid: 0x2bc5, name: USB Receiver
[libretro INFO] [MAME 2003+] Preparing emulated CPUs for execution.
[libretro INFO] [MAME 2003+] arkanoid hiscore memory map found in hiscore.dat!
[INFO] [Environ]: SET_GEOMETRY: 256x224, Aspect: 0.750.
[libretro INFO] [MAME 2003+] loading arkanoid.hi hiscore memory file...
[libretro INFO] watchdog armed
[INFO] [Config]: Saving salamander config to: "sd:/retroarch/retroarch-salamander.cfg".
[INFO] [Config]: Saved new config to "sd:/retroarch/retroarch.cfg".
[INFO] [Core]: Content ran for a total of: 00 hours, 00 minutes, 15 seconds.
[INFO] [Runtime]: Saving runtime log file: "sd:/retroarch/cores/playlists/logs/MAME 2003-Plus/arkanoid.lrtl".
[INFO] [Core]: Unloading game..
[libretro INFO] [MAME 2003+] saving arkanoid.hi hiscore memory file...
[INFO] [Core]: Unloading core..
[INFO] [Core]: Unloading core symbols..
[INFO] [Core]: Saved core options file to "sd:/retroarch/config/MAME 2003-Plus/MAME 2003-Plus.opt".
[INFO] [Video]: Does not have enough samples for monitor refresh rate estimation. Requires to run for at least 4096 frames.
[INFO] [Video]: Does not have enough samples for monitor refresh rate estimation. Requires to run for at least 4096 frames.
```
I think the most important part is `[INFO] [Environ]: SET_GEOMETRY: 256x224, Aspect: 0.750.` If I'm reading it right, the `Aspect: 0.750` is correct, but the resolution being passed to RetroArch is un-rotated.

mahoneyt944:
What setting are you using for `settings->core->allow rotation.` you'll need show advanced settings enabled under settings->user interface

vaguerant:
Settings -> Core -> Allow Rotation is set to OFF. Just for reference, I'm on libretro/RetroArch@06a2367, which is five hours old as of this comment. MAME 2003 Plus is at 379c0e9, a day old. As far as I can tell, I have everything set up correctly, and the logs seem to support that. Quoting the log from above:
```
[INFO] [Environ]: SET_ROTATION: 0
[libretro INFO] [MAME 2003+] This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally
[libretro INFO] mame internal: video_flip_x:1 video_flip_y:0 video_swap_xy:1 video_hw_transpose:1
```
Everything seems to be working as expected based on the log, with the one exception of the geometry.

grant2258:
this will mess up ra rotations

grant2258:
This should be where you want to be on both to do rotations. There will be shader issues with vertical games when RA cant handle them internally. 
```
diff --git a/src/mame2003/video.c b/src/mame2003/video.c
index 483f2915..8fd8fd24 100644
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -32,6 +32,7 @@ unsigned video_do_bypass;
 unsigned video_stride_in, video_stride_out;
 bool video_flip_x, video_flip_y, video_swap_xy;
 bool video_hw_transpose;
+bool libretro_rotation_failed=0;
 const rgb_t *video_palette;
 uint16_t *video_buffer;
 
@@ -58,8 +59,17 @@ void mame2003_video_get_geometry(struct retro_game_geometry *geom)
    geom->max_width = geom->max_height = max_dim;
 
    /* Hardware rotations don't resize the framebuffer, adjust for that */
-   geom->base_width = video_hw_transpose ? vis_h : vis_w;
-   geom->base_height = video_hw_transpose ? vis_w : vis_h;
+   if (!libretro_rotation_failed) 
+   {   
+      geom->base_width = video_hw_transpose ? vis_h : vis_w;
+      geom->base_height = video_hw_transpose ? vis_w : vis_h;
+   }
+
+   else 
+   {
+      geom->base_width = vis_w;
+      geom->base_height = vis_h;
+   }
 
    geom->aspect_ratio = video_hw_transpose ? (float)video_config.aspect_y / (float)video_config.aspect_x : (float)video_config.aspect_x / (float)video_config.aspect_y;
 
@@ -137,12 +147,18 @@ void mame2003_video_init_orientation(void)
       }
 
       else
-        log_cb(RETRO_LOG_INFO, LOGPRE "This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally\n");
-
+      {
+         log_cb(RETRO_LOG_INFO, LOGPRE "This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally\n");
+         libretro_rotation_failed=true;
+      }
    }
+
    else
+   {
      log_cb(RETRO_LOG_INFO, LOGPRE "RetroArch does not support this type of rotation, using mame internal rotation instead\n");
-
+     libretro_rotation_failed=true;
+   }
+   
    tate_mode = options.tate_mode;
 
    /* Set up native orientation flags that aren't handled by libretro */
```

vaguerant:
Can confirm that works perfectly over here, fixes both Custom and the 1:1 PAR aspect ratios, since they both rely on the geometry, as well as Core provided since that relies on aspect ratio.

grant2258:
any vertical flips ra cant handle will mess up shader on all platforms unfortunately RA  doesnt support flip_x and flip_y feel free to put a pull request in with this if you want to add it. I only use aspect ratios as anything else doesnt really fit well to crt disiplays. No need for credits and such if you think its useful add it or change as you see fit.

grant2258:
just checked fbneo it seems to not update the x/y rotations properly and it doesnt rotate at all with vertical mode on only checked with circus charlie. Might be worth getting that fixed for wii u users as well or at least bring it to the core devs attention can be fixed doesnt mean they will though.

mahoneyt944:
@grant2258 not related to this core but noticed mame2010 has rotation issues with some games too. Can't recall which ones I ran but I'm sure it has to be related to ra vs. core rotation handling.

vaguerant:
I suppose the subset of platforms that have a) no working rotation and b) shader support is probably quite small. Rotation works on almost everything, the only platforms I know with broken rotation are Wii U and 3DS. As far as I know, RetroArch on 3DS doesn't support shaders at all (not much point on a 240p screen). Ultimately, I'd say rotating correctly is more important than supporting shaders correctly.

grant2258:
@vaguerant I agree rotation has to be done for usability. Even with ra rotations there are some games that will have the shaders wrong like kick but this is because its doesnt support all rotations needed. 

@mahoneyt944 mame2010 needs lots of work no one seems interested though if it had maintainers it might track so motion in getting updates . I guess people would migrate to fbneo or mame current with stronger hardware anyway in all honesty

vaguerant:
Just for reference, libretro/RetroArch#13353 (just merged) fixes frontend rotation on 3DS and Wii U, so this fallback rotation mechanism is no longer needed on those platforms.

grant2258:
Not sure if there are any other platforms needing it wont do any harm keeping it though. Good job in fixing the issue on the frontend/backend side.

vaguerant:
Still on this geometry issue though, I'm trying to understand the purpose of what's going on here:

1) Swap geometry in `vis_height` and `vis_width`:
https://github.com/libretro/mame2003-plus-libretro/blob/4bb681c1b71de4f2ac018d671b6b6a1cdc7201d9/src/mame2003/video.c#L76-L81

2) Populate the (swapped) geometry from `vis_height` and `vis_width` into `vis_h` and `vis_x`
https://github.com/libretro/mame2003-plus-libretro/blob/4bb681c1b71de4f2ac018d671b6b6a1cdc7201d9/src/mame2003/video.c#L53-L54

3) Swap geometry in `vis_h` and `vis_w`:
https://github.com/libretro/mame2003-plus-libretro/blob/4bb681c1b71de4f2ac018d671b6b6a1cdc7201d9/src/mame2003/video.c#L60-L62

Is there a situation where `video_swap_xy` is hit but `video_hw_transpose` is not? Because otherwise this looks like you're just swapping the geometry then swapping it back again, meaning the geometry gets borked for any vertical game. This lines up with my experience of using 1:1 PAR or Custom aspect ratios giving the opposite of the expected geometry when running vertical games.

On my end, I replaced L61-62 with this and everything seems to work perfectly:
```
   geom->base_width = vis_w;
   geom->base_height = vis_h;
```
Is there something this breaks? Everything seems to work as expected on my end. Horizontal, vertical and even upside-down games all report their geometry correctly with this change.

grant2258:
You have to cancel the mame rotation when ra does it that will null swapxy

grant2258:
to be honest I have no idea what Ra is doing in the background if you just want the aspect ration changed and keep the xy that is easy enough im pretty sure that will mess up core provided aspect ratio though

grant2258:
It would be preferable if you could add something like this so if things change we can just reintroduce it, Ra really should be update the xy imho it sticks and stones though this does get confusing when your dealing with two rotations, ps RA cant handle kick make sure thats working fine for you as well. 
```
diff --git a/src/mame2003/video.c b/src/mame2003/video.c
index 483f2915..f25c0ed4 100644
--- a/src/mame2003/video.c
+++ b/src/mame2003/video.c
@@ -32,6 +32,7 @@ unsigned video_do_bypass;
 unsigned video_stride_in, video_stride_out;
 bool video_flip_x, video_flip_y, video_swap_xy;
 bool video_hw_transpose;
+bool libretro_rotation_failed;
 const rgb_t *video_palette;
 uint16_t *video_buffer;
 
@@ -58,11 +59,22 @@ void mame2003_video_get_geometry(struct retro_game_geometry *geom)
    geom->max_width = geom->max_height = max_dim;
 
    /* Hardware rotations don't resize the framebuffer, adjust for that */
-   geom->base_width = video_hw_transpose ? vis_h : vis_w;
-   geom->base_height = video_hw_transpose ? vis_w : vis_h;
 
-   geom->aspect_ratio = video_hw_transpose ? (float)video_config.aspect_y / (float)video_config.aspect_x : (float)video_config.aspect_x / (float)video_config.aspect_y;
+  /* ra doesnt want updated rotated geometry remove condtion if this changes
+   if (!libretro_rotation_failed)
+   */
+   if (0) 
+   {
+      geom->base_width = video_hw_transpose ? vis_h : vis_w;
+      geom->base_height = video_hw_transpose ? vis_w : vis_h;
+   }
+   else
+   {
+      geom->base_width = vis_w;
+      geom->base_height = vis_h;
+   }	
 
+   geom->aspect_ratio = video_hw_transpose ? (float)video_config.aspect_y / (float)video_config.aspect_x : (float)video_config.aspect_x / (float)video_config.aspect_y;
 }
 
 void mame2003_video_update_visible_area(struct mame_display *display)
@@ -112,6 +124,7 @@ void mame2003_video_init_orientation(void)
    unsigned rotate_mode;
 
    rotate_mode = 0; /* Known invalid value */
+   libretro_rotation_failed=0;
    tate_mode = options.tate_mode;   /* Acknowledge that the TATE mode is handled */
    video_hw_transpose = false;
 
@@ -137,12 +150,18 @@ void mame2003_video_init_orientation(void)
       }
 
       else
-        log_cb(RETRO_LOG_INFO, LOGPRE "This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally\n");
-
+      {
+         log_cb(RETRO_LOG_INFO, LOGPRE "This port of RetroArch does not support rotation or it has been disabled. Mame will rotate internally\n");
+         libretro_rotation_failed=true;
+      }
    }
+
    else
+   {
      log_cb(RETRO_LOG_INFO, LOGPRE "RetroArch does not support this type of rotation, using mame internal rotation instead\n");
-
+     libretro_rotation_failed=true;
+   }
+   
    tate_mode = options.tate_mode;
 
    /* Set up native orientation flags that aren't handled by libretro */
```

vaguerant:
I'm not at a computer right now to be able to get a log or similar, but this screenshot shows where the problem is:
19xx

The game running here is 19XX. The hardware's native resolution is 384\*224. For a vertical game like 19XX, that obviously should translate to 224\*384. However, if the aspect ratio is set to 1:1 PAR or Custom, the geometry RetroArch is providing to the user is the original 384\*224. You can see that the game running in the background looks incorrect (very wide instead of very tall), if you switch to Custom and Integer Scale, it's offering integer scales of 384\*224, not 224\*384.

The Core Provided aspect ratio is fine; I don't think the geometry and desired aspect are actually tied together, are they? Changing the geometry should only change the geometry, not the aspect ratio.

I can barely read code, so I can't really be helpful in understanding why it works this way,  vertical games all have their horizontal geometry. Setting the aspect ratio to Core Provided "resolves" this issue because the geometry just isn't used when you run with Core Provided aspect ratio, it only matters when using 1:1 PAR or Custom.

grant2258:
Well RA need to be a bit more clear it does not report x/y changes when rotated which it should do to be honest. Check the statistics youll see if your using non Ra rotation you need to feed it the x/y which is swapped in the code for PAR to work the code does this in mame with swapxy.

screen shot of mappy with par and stats with no libretro rotation
Screenshot from 2021-12-17 02-16-46

screen shot of mappy with par and and stats with no libretro rotation
Screenshot from 2021-12-17 02-20-45
ats with RA rotation

This is with the new code as you can see both use different xy RA isint updating the xy stats when rotating which is strange the new code works but it defo something that needs looked at. Looks like Ra want the raw image xy and it doesnt update the xy stats when it does rotation causing the confusion..

grant2258:
I have to be honest I dont see any real use for these custom scaling as crts dont fit a 1:1 pixel to crt dot check out sf2 look at squares and circles in particular once you see it you cant unsee it. But if people want it should work I guess feel free to make any changes just fix anything that gets broken if it happens the patch should be fine.

mahoneyt944:
Would it be less hassle to always let ra do the rotations and any platform that doesn't work properly can be filed through RA issues? Letting mame do it fixes the issue in the short run but it's just masking the issue if RA can't and should be fixed in the frontend.

grant2258:
Ra cant do all the rotations I doesnt support them all

mahoneyt944:
Which ones can't it do?

grant2258:
swapxy flipx and flipy

grant2258:
@mahoneyt944 there is no issue here apart from ra not being clear on what it doing or it updated the xy without showing it on the stats the video code isint big its all done there

mahoneyt944:
Need to make a issue for that I guess

vaguerant:
@grant2258 There's a handful of cases where 1:1 or Custom are useful. On 3DS, the 400\*240p screen is too low for any scaling to not look awful. I do make an exception for CPS games just because 384 is very obviously too wide, but if it's something like a game running at 256*240, you're better off with 1:1 on a 240p screen rather than using Core Provided to get a slightly more accurate image shape that looks atrocious.

Then on Wii U, you have a situation where shaders are broken, so it's just not possible to get a Core Provided image that looks good--you'll have pixel-shimmering all over the image if you use any non-integer scale, unless you just bilinear the whole thing to hell. On other platforms, you could use something like the sharp-bilinear shader to avoid the shimmer and the blur, but on Wii U you just have to take your pick. Or, you can use a custom integer scale. For CPS games, you can even use Custom to do a 3x horizontal/4x vertical integer scale for a final output image of 1152\*896, which gives you an image that's only about 3% slimmer than the Core Provided 4:3, but doesn't have any scaling artifacts.

grant2258:
@vaguerant Im just speaking from my perspective the patch will fix both the stats will confuse users that know what they are doing if they look wrong is all just a live with case scenario. These are honestly not options I have used in RA I did fix this a while back but i undid it because I didnt leave a note about it when I done some changes thats why i added the comment and if condition on the patch. Just test othunder kick and dotron to make sure the scaling works properly with non supported RA rotations

vaguerant:
Re-reading through the thread, I agree that this is where all the confusion comes from:
> Well RA need to be a bit more clear it does not report x/y changes when rotated which it should do to be honest.

>This is with the new code as you can see both use different xy RA isint updating the xy stats when rotating which is strange the new code works but it defo something that needs looked at ra end to not confuse whats going on. Looks like Ra want the raw image xy and it doesnt update the xy stats when it does rotation causing the confusion.

RetroArch *always* expects the core to be in charge of geometry. That's the intended behavior, which I've confirmed with jdgleaver from libretro and barbudreadmon from Team FBNeo. Certainly, you can make the case that RetroArch *should* be handling geometry changes when they are caused by rotations, but as far as I can tell there's no appetite for that on RetroArch's side; they prefer that the core submits any geometry and/or aspect ratio changes as needed.

The only (hypothetical) upcoming change I'm aware of on the RetroArch side (per jdgleaver) is having RetroArch handle aspect ratio changes **only if the rotation is called from within RetroArch** (i.e. via the Settings > Video > Output > Video Rotation option), since that's a situation where the core doesn't "know" about the rotation and can't submit any updates for it. The RA policy is that if the rotation comes from the core, it's the core's job to handle any associated aspect ratio and geometry changes.

mahoneyt944:
Viewing patches from my phone is difficult for me. If there's a good solution I'll be happy to merge the PR once ready.

Ideally we should integrate with RA as much as possible, when possible I should say... and allow the core to handle case scenarios when not, as grant pointed out with flips and swaps etc. Whatever is best for long-term friendliness to the code.

barbudreadmon:
>Not sure if there are any other platforms needing it wont do any harm keeping it though

Afaik it'll prevent `Settings > Core > Allow Rotation` from working as expected (rotation will happen even after user forced it off ?). There are also [other reasons](https://github.com/libretro/mame/issues/261) you might want to avoid software rotation altogether in a libretro core's code.

vaguerant:
It's worth noting that it's probably still needed for some users on Wii U. There's an issue where the RetroArch Ozone GUI menu selection bar rotates whenever libretro rotation occurs which affects some users. I used to be one of those users, but then the bug just went away for me, but others still encounter it. I've tried everything I can think of to reproduce the issue but can't get it to happen.

grant2258:
them bugs are the worst we do need the rotation for teh one RA cant handle as well just mirroring

PR #1209, opened 2021-11-29
Undrfire crosshairs and cleanups

mahoneyt944:
This was never actually finished and was later replaced by newer methods of displaying the sights. This is my current wip to track the input which uses a large range noted in undrfire.c driver.


mahoneyt944:
@grant2258 this is how the game uses the input apparently the original used a large range. Might be a way to use the same in the crosshair tracking. Or even move it into that function?
https://github.com/libretro/mame2003-plus-libretro/blob/3aaf473203f2d4ae22a06d0b88202d24c42d1380/src/drivers/undrfire.c#L405-L438

grant2258:
@mahoneyt944 could your try if everything still works on full aspect ratio on android so the image stretches.

mahoneyt944:
@grant2258 tested the tracking. It is the same in both views...core provided and full. There wasn't any crosshairs before since it wasn't implemented yet for undrfire. So at least its pretty close now and working via core option like the others, I'll look at it more some other time.

grant2258:
@mahoneyt944 cheers for that wanted to see how the android port was dealing with scaling so udev matches as far as line of site guns go.

PR #1227, opened 2021-12-04
Tag player number draw_crosshair

mahoneyt944:


grant2258:
This isn't ideal for all games if your shooting thing going past at the same position. You just timing the trigger you'll also need to check any player buttons for activity in this as well to not cause unexpected side effects.

PR #1251, opened 2021-12-21
Isolate ost support part 2

mahoneyt944:
Ost samples are now controlled externally, and flag the driver to play default sounds when required. Most of the repetitive code is now isolated into separate functions.  Found / fixed some minor issues while refactoring. Removed some variables that are no longer needed with the new structure. Hopefully this makes it easier to maintain and update in the future.


grant2258:
https://github.com/libretro/mame2003-plus-libretro/blob/4890b10822ef2a2a5b8ed95574e0ec73ea615be3/src/ost_samples.c#L311-L317 is incorrect behavoir for this core to stop samples we dont stop by sample number we stop by channel.

mahoneyt944:
This is how the original code was working. Maybe was hooked up wrong all along.
 https://github.com/libretro/mame2003-plus-libretro/blob/c10641514f5acfda34c74a2c8962a75526148e6c/src/sndhrdw/williams_sndhrdw.c#L840-L947

Could test with channel only.

grant2258:
that the acase for nba jam yea i know i fixed this up on my port think ff was ok though would need to check the source but it is incorrect not sure whats been fixed here and whats not.

mahoneyt944:
Should be as easy as updating the new ost_stop_samples function and removing another uneeded variable "sa_count".

Issue #290, opened 2018-06-14
[Bounty] correct Donkey Kong color palette 


barbudreadmon:
### [Bounty contributions make a difference - please donate at this link today to support this fix](https://www.bountysource.com/issues/59677980-correct-donkey-kong-color-palette)

dkong colors got fixed in lr-fbalpha (they were wrong before mame141u2, and fba used old code too) : https://github.com/libretro/fbalpha/commit/a949908555ad1e558b9b2161bbee6e49d6ee2a21

Perhaps you can somehow port this to mame2003-plus.

Best regards.

arcadez:
Firstly thanks for the headsup, i see this original commit then a further two afterwards which i
assume are connected with it.......

https://github.com/dinkc64/mame/commit/10ad09bac8b399bd817238f9c09f17a0f84615b6

Whether or not the palette and resnet changes would be backportable to this ole core will
need some investigation as from experience i'll bet how we handle this graphical code in 
MAME2003 will be totally different there's no resnet.c for starters.

Regards

Wilstorm:
@arcadez - Off topic I have to comment here real quick because I've been thinking about you these past few days since the Summit and when you made comment that you wanted to get a things done before the next World War. Here's to hoping! ;)

markwkidd:
Because I'm interested in looking at the Pauline version maybe I can look at this at the same time.

arcadez:
Two problems firstly in MAME78 dkong.c doesn't use resnet so would need to be updated,
second problem our resnet is nothing like the versions in later MAME so we'd likely have to
update that and all the drivers that use it.

Ya see how an easy fix in later MAME is not exactly as simple for this core.!!

barbudreadmon:
Those are the colors after the fix : 
![dkong](http://adb.arcadeitalia.net/media/mame.current/ingames/dkong.png?release=185)
Colors prior to 0.141u2 were more red, i guess it is the case in mame2003-plus

3 games (and their clones) should be affected by this change : 
> 0.141u2: Couriersud fixed a bug in palette_normalize_range which caused color distortion. The luminance normalization now converts r,g,b to y,u,v and normalizes y prior to converting back to r,g,b. This affects e.g. Radar Scope, Donkey Kong and Mario Bros.

The fix in fba was ported from MAME.

grant2258:
thanks for the info is appreciated @barbudreadmon

markwkidd:
[...]

PR #1267, opened 2022-01-06
imperfect color flag for dkong romsets

barbudreadmon:
Follow-up of https://retropie.org.uk/forum/post/271428 and https://github.com/libretro/mame2003-plus-libretro/issues/290. Not sure there are plans to ever fix this since the current colors are popular in their own way, so the imperfect colors might as well be properly documented


mahoneyt944:
Checks have failed. Please use the proper syntax.

barbudreadmon:
Oops, i didn't notice you used a slightly different macro, it should be ok now

PR #1270, opened 2022-01-08
Street Fighter 2 ost sample support

mahoneyt944:
- removed duplicate "gameover" samples 90/91 and corrected with 50/51
- rewrote volume fade controller for simplistic design
- re-hooked fading triggers, should fix fading on all endings, and prevent fade loops
- found a dispute in case 0x11 (gameover and credits screen), original sound loops on the special credits screen but the ost sample ends with a fade which sounds a little odd looping....however not looping means a silent credit screen. I decided to loop for now. We may want to use the default sound here instead since it fills both conditions better.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -215,7 +215,6 @@ android-arm64-v8a:
     - .core-defs
   only:
     - master
-    - Sf2
 
 # Android 64-bit x86
 android-x86_64:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -215,6 +215,7 @@ android-arm64-v8a:
     - .core-defs
   only:
     - master
+    - Sf2
 
 # Android 64-bit x86
 android-x86_64:
--- a/src/drivers/cps1.c
+++ b/src/drivers/cps1.c
@@ -102,9 +102,15 @@ static WRITE16_HANDLER( cps1_sound_command_w )
 	}
 	*/
 	
-	/* We are playing Final Fight. Let's use the samples.*/
+	/* We are playing Final Fight. */
 	if(ff_playing_final_fight && options.use_alt_sound) {
 		if(generate_ost_sound_ffight( data )) {
+			if(ACCESSING_LSB) soundlatch_w(0,data & 0xff);
+		}
+	}
+	/* We are playing Street Fighter 2. */
+	else if(sf2_playing_street_fighter && options.use_alt_sound) {
+		if(generate_ost_sound_sf2( data )) {
 			if(ACCESSING_LSB) soundlatch_w(0,data & 0xff);
 		}
 	}
@@ -158,6 +164,9 @@ static INTERRUPT_GEN( cps1_interrupt )
 	/* works without it (maybe it's used to multiplex controls). It is the */
 	/* *only* game to have that. */
 	cpu_set_irq_line(0, 2, HOLD_LINE);
+
+	if(sf2_playing_street_fighter && options.use_alt_sound)
+		ost_fade_volume();
 }
 
 /********************************************************************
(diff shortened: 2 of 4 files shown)


kontinik:
Is out there a samples source?

mahoneyt944:
@KMFDManic was hosting them at one point. Idk if there are sources still available or not.

kontinik:
I've checked certain google drive folders and I couldn't find them. SF2 reminds my childhood

KMFDManic:
@kontinik ill repost them when I get a chance to. I had to temporarily take them down from my google drive when changing data plans.  I was stuck on a higher plan and they wouldnt let me downgrade. So, simply canceled and resubscribed to 30 USD a year versus the 300 USD a year they wanted heh.

Wilstorm:
@KMFDManic - I think I am missing a few also. I wasn’t sure of the reason why they came down but I figured I would probably come across some at some point. I’ll have to keep an eye out… 

Thanks for the bootstraps… I keep all the goodies and extras you guys create… I have a few of them for shooters to facilitate them working properly..

I know that wasn’t the issue you were working on, but it got me thinking of the bootstraps when I was reading the other issue… It’s all good stuff!!! 😎

grant2258:
> @kontinik ill repost them when I get a chance to. I had to temporarily take them down from my google drive when changing data plans. I was stuck on a higher plan and they wouldn't let me downgrade. So, simply canceled and resubscribed to 30 USD a year versus the 300 USD a year they wanted heh.

sadly another example of great customer loyalty but the big corps its the same over here with car insurance, breakdown cover basically anything only new customers get any incentive to stay. I got rid of the car this year insurance is joke over here it working out cheaper bussing it these days a sad reflection of the times though. Ill probably get pissed off and end up paying it though because there nothing more convenient than a car unfortunately. Lets see how long my will power lasts haha. Im taking a stand for moral reasons, I needed to get a new car anyway will get something smaller this time like a wee picanto or something. 

Dont even get me started on the power companies smart meter that dont work. Have had 2 new gas and electric meters put in and they still dont work well the electric does now but the gas aint having any of it. Have to wait for a new gas meter and not so smart meters to be attached to it yet again. I think internet archive might have them samples they used too anyway.

Wilstorm:
@grant2258 - I dig it! You just say it straight up like it is… Thanks for the tip on the Internet archive. I will do a poking around! 

Have a good weekend.. we don’t have a lot planned but hanging around the house because it’s nothing but cold here, this morning going to work it was 21 below Fahrenheit which is like 29.5 below in Celsius. We’re just getting killed with this cold weather in the upper Midwest but only one snowstorm this year… Which is very strange weather for us… Not the cold so much but the lack of snow..

KMFDManic:
@kontinik @Wilstorm Looks like the only posted ones are for mortal kombat, final fight, double dragon.  I have many more than are posted, such as Ikari Warriors, and so on.  Some are coded into Plus, at least the most stable ones.  Xtreme has several more, such as Ghosts 'n Goblins, Shinobi, Joe and Mac, Haunted Castle, and so on, that are for better or worse workable.  I enjoy them all:)  I will have to do an audit of the ones supported in Plus/Xtreme, and upload them to my Google Drive, again.   

Some samples are incomplete...as in, they are "supported", but the samples aren't fully made for them yet!  I believe Xtreme has almost 2 dozen supported "sample" games, now...Mahoneyt944 and Grant2258 were quite helpful, as well as kurrono, gpstar81, big blue frontend, etc:)  

I will try to get things audited as soon as I can, so we know which need some samples fixed up...And, anyone willing to help, can help complete some of the sample sets, of course:)  Just have to be creative enough to make ones that wouldn't irritate you to kingdom come on the 100th playthrough!  The whole idea is to try to have fun ones that can go head to head with the originals...MoonWalker, for example...Absolutely love using the "real" songs for that one...never ever gets old!  Also, great with the Genesis version!:)

PR #1295, opened 2022-02-26
PlayChoice-10 Joypad Support

KMFDManic:
PlayChoice-10 much easier to play, if actually supported by a controller:)

Select=Select
Start=Start
Enter=Select (Top Screen Information)


KMFDManic:
@mahoneyt944 @arcadez2003 @grant2258 

I decided to run some PlayChoice-10 games and discovered they are not optimal, especially for those with JUST NES controllers.  I updated the inputs, and personally tested.  Select will toggle between the information screens on top, as well as "add additional time", act as the in game select button, per normal.  Start acts as it should.  These games run surprisingly well, with seemingly little or no slowdown...comparatively speaking to other NES emulators I have run these on with lower spec Minis.  

"Free Play" might be something to consider implementing.  I just wanted to at least handle the important functions, for the time being.  Other than that, I may adjust the "analogs", so Duck Hunt is more controllable with a controller, as well.  But, this is a great start, so people can jump right into these games without any need of Mame Menu Remaps.

PR #1309, opened 2022-03-21
WIP DO NOT MERGE - Rename umk3p to umk3pb1, and add umk3pb2

palmerj:
Adds Ultimate Mortal Kombat 3 Plus Beta II: https://www.mortalkombatplus.com/games/ultimate-mk3. Thanks MKPlus team - TehDrewsus and ZPaul2Fresh8!!

This PR keeps the existing beta 1 for people that still wish to play that version.


mahoneyt944:
Did not know this, replied on the forums with further details.

palmerj:
All good, it's pretty much the same as your PR, with the exception of the setup of the NRAM file. I also updated the history.data and catver.ini. Can you also do that?

In general any reason why the NRAM files are not setting freeplay? Seems like a good thing to do in the context of retroarch.

mahoneyt944:
Freeplay is never set as default in any case I've seen so I don't force it on anyone. Not all games support Freeplay so you'd end up with a mix of games with and without it which would be confusing to remember which ones do and don't.

One may argue that the select button for inserting digital coins is emulated Freeplay 😂

I can add those other additions unless you want to open another PR and I'll merge.

mahoneyt944:
Updated everything except history.dat it's too large to edit by mobile I get it when I'm home later. Also added credit for you in the changelog where beta 1 was originally added.

PR #1310, opened 2022-03-22
Support Ultimate Mortal Kombat 3 Plus Beta 2

mahoneyt944:
- Rerom umk3p to umk3pb1 and umk3pb2 respectively.
- update bootstrap for umk3pb1
- new bootstrap for umk3pb2


palmerj:
Did you see my earlier PR here: https://github.com/libretro/mame2003-plus-libretro/pull/1309 and note that the dev of the hack didn't want it merged yet?

Zoltan45:
the game should be named umk3p as well.  it is different from umk3 the same way umk3 is differnet from mk3.  umk3p should be  the current rom name.  umk3p and umk3pb1 would be more appropriate, and probably mame accurate.

Zoltan45:
> Did you see my earlier PR here: #1309 and note that the dev of the hack didn't want it merged yet?

it seems the cat is out of the bag.  Just let it fly now.

mahoneyt944:
Back when multiple RetroArch cores began adding umk3p beta 1 support, mame2003-plus was the only one that did not use umk3pb1 as it's naming scheme. This was actually a hassle if you heavily test the game in different emulators... you'd have to rename the zip before cross comparing the emulation. Right or wrong idk, but using b1 and b2 designations is more future proof and probably what other RA cores will use. Technically beta 2 is the last release to the public, but that does not mean it's the final revision / release which would get designated umk3p. I'm surely not attached to any naming scheme one way or the other, but internally (such as the bootstrap, or init functions) we use the b2 designations so I was just keeping it consistent in that regard.

Zoltan45:
I'm trying to convince you to use naming as I think mame would, but you are free to do as you wish.  The current version would have the main game name umk3p and beta 1 would be umk3pb1.  If/when a beta 3 comes out.  umk3p would then become umk3pb2 and beta 3 would become the new umk3p.  I understand where you are coming from, I am just trying to keep things in line with how it would be, I believe, if it gets added into mainline mame.  For example, when umk3 revision 1.0 and 1.1 were the only versions in mame 1.1 would have been named umk3.zip and 1.0 would have been umk3r10.zip and then when 1.2 got dumped, 1.2 became umk3.zip and umk3 r1.1 went from being umk3.zip to now being umk3r11.zip.  PS I'm not saying this happened(it didn't).  Just a relatable example.  In the end, maybe we're both right?

palmerj:
I think it's six of one, half a dozen of the other. I also promote the b1/b2 naming convention because other emulators have already done this. e.g. FBNeo. Also, having each version there allows you to play past beta versions. But in the long run, I understand it adds a little more complexity. HBMAME also has the plus hacks added, but its naming structure is pretty out there and not something this project should copy.

Zoltan45:
theres never gonna be a one size fits all approach with mame lol

Zoltan45:
fbneo added the game as umk3p.zip lol

mahoneyt944:
They also dropped support for beta 1 all together. Possibly because beta 1 is not available to download now? Idk.

mahoneyt944:
This is actually more complicated than just a rename, if we rename beta 2 to umk3p and drop beta 1 support than the cfg and nvram file from when beta 1 was called umk3p will cause a conflict with the new files. This means any user that used beta 1 and doesn't manually delete it will load a bad nvram file which could cause a whole bunch of glitches

Zoltan45:
> This is actually more complicated than just a rename, if we rename beta 2 to umk3p and drop beta 1 support than the cfg and nvram file from when beta 1 was called umk3p will cause a conflict with the new files. This means any user that used beta 1 and doesn't manually delete it will load a bad nvram file which could cause a whole bunch of glitches. But we also shouldn't support beta 1 if it's no longer available for download. It would be hard to obtain to those building romsets.

the game is still available in the same form it always has been, and I also understand your concerns.

I would also hope that no one is ever under the illusion that working with emulators in any way is ever simple and has no hiccups like that.  I am not even referring to this issue specifically.  I'm just talking in emulation generally.  There's always something (to go wrong), isn't there?

 I appreciate the goal to get to make it as easy as possible though. So now what is the name going to be in libretro now? umk3p as here or umk3pb2 as in libretro/FBNeo?

mahoneyt944:
Well, to be consistent with my original statement having conformity makes life easier and being that I cannot find links to beta 1 on the official site for this hack, I don't see it as feasible for those who are building their own romset to find it. I'm sure it's hosted elsewhere on the web, but the creators agreement asks that no one share the file anywhere else so that there is a centralized source for the hack. So I'll be dropping support for beta 1 and renaming Beta 2 to umk3p.

palmerj:
Again I don't think there is a clear answer here yet as it still won't be consistent across all main emulators - Fightcade will still be umk3b2.zip. If that changes then ok 100% swtich.

Regarding the beta1 patches not being online anymore, I hardly think that's an issue as communities always make sure things are available. If that weren't the case most people wouldn't have any games to play ;)

mahoneyt944:
We will not be supporting beta 1 anymore to comply with the creators wishes. They do not host the earlier revision to download, and it should not be shared by anyone else per the download agreement.

```
2. You will NOT share this file. It will always be available on this website and I’d prefer to keep it in one location for statistical purposes.
```

Zoltan45:
Fightcade does not have anything to do with libretro.
  The patches for beta 1 should be available again. That’s a whole other topic though. And when that gets resolved is anyones guess.
 I would also guess that since the game was added to mainline fbneo that fightcade will now change theirs to match.


Sent from Yahoo Mail for iPhone


On Thursday, March 31, 2022, 1:17 AM, Jeremy Palmer ***@***.***> wrote:




Again I don't think there is a clear answer here yet as it still won't be consistent across all main emulators - Fightcade will still be umk3b2.zip. If that changes then ok 100% swtich.

Regarding the beta1 patches not being online anymore, I hardly think that's an issue as communities always make sure things are available. If that weren't the case most people wouldn't have any games to play ;)

PR #1311, opened 2022-03-23
More updates

mahoneyt944:


Zoltan45 (commented):
Ultimate Mortal Kombat 3 + is a new game in the bloodline the same way Ultimate Mortal Kombat 3 is a new game compared to Mortal Kombat 3. Beta 2 should be called umk3p as current versions take the game's main name.

PR #1373, opened 2022-06-10
Initial video update to system16

mahoneyt944:
@MistyDreams. I think this is a pretty safe first step. I added the sprxoffset shift for the new video code and then hooked up tturf, tturfu, and tturfbl. And the video is fixed. This obviously doesn't cover any system16a stuff you're working on but I don't want master to get too far behind so we can test incrementally if we find new issues down the road.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -215,6 +215,7 @@ android-arm64-v8a:
     - .core-defs
   only:
     - master
+    - Mist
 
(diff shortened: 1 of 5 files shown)


MistyDreams:
can do what you like on your repo dont really mind im still working away so things will change

mahoneyt944:
Ok merging initial wave. More commits to be had.

PR #1381, opened 2022-06-18
New sys16 for 2003plus repo

MistyDreams:
new system16 driver doesnt have ttilemap.h and tilemap.c changes. Does not effect current old system16 driver. Users the old sound core will need the new one for samples to play on cotton. @arcadez2003 here it is will save you the hassle


MistyDreams:
@mahoneyt944 not sure what you guys are doing with tabs n spaces all seems mixed at the moment applied the repo editorconfig settings to my driver seems its not consistent among files do you want  tabs or spaces?

mahoneyt944:
It's not super important which you use. I just try to keep new additions somewhat consistent. So which ever spacing you like personally, but try to not to use alternating spacing within the same function or whatever so each section reads easier.

As far as this PR, looks pretty good to me. Some things we might want to decide on before we pull the trigger though is which games we want to use and where. Newer mame places the official roms in segas16b.c and the bootlegs in system16.c we kind of have the opposite going.

Do we want to follow this designation for this core or not worry about it?

Also, src/vidhrdw/segas16a.c what's going on with this? Is it being used in this PR or for a future addition? It's not in the makefile.

MistyDreams:
nothing going on with system16a at the moment can be plugged in if needed or removed. I dont mind what you guys do long term where any games go. That's beyond the scope of this pr little bits at a time you said. Long term that will be your choice.

I just put the games I added in this update is needed for cotton. The sound core is for another for pr though so can be addressed at a later date it has been tested. The branch i have at the moment isint good for a pr was a testing branch.

mahoneyt944:
Do you feel this is ready to merge or are you still tweaking things?

MistyDreams:
no its done apart from any changes you want done to whats there already.

PR #1420, opened 2022-08-05
Sf1 OST

mahoneyt944:
@KMFDManic here's the converted version.


KMFDManic:
Exceptional turnaround:) Looking very nice with the house cleaning! There will be more Custom OSTs incoming in the near future.

mahoneyt944:
@KMFDManic thanks. I know there's a little extra stuff in this commit than just sf1 but the baseline on the new style is each OST supported game is confined to a external function which handles all the custom sound mixing. All the code is isolated to ost_samples.c so the code in the driver itself is just simple hooks. It's very simple and easy to work with. The biggest benefit is we are reusing all the common code, so this significantly reduces the "if, else if, and flag" spaghetti for each hookup.

PR #1426, opened 2022-08-10
add hyper street fighter 2 and clones for arcadez2003

MistyDreams:
@arcadez2003  This is an example of how to add the encrypted games


MistyDreams:
Removed the need to ROM_REGION16_BE( CODE_SIZE, REGION_USER1, 0 ) set in the rom loads. We are just using automalloc now. So basically to roms you just copy the mame ones and change the regions to ours and add the init. Ive added all the 19xx and hsf2 sets.

arcadez2003:
Thanks

PR #1427, opened 2022-08-10
add this as requested

MistyDreams:
Not sure why this is needed but is added per your request controls still work with and without it.


mahoneyt944:
These are control names for RetroArch.

PR #1428, opened 2022-08-11
add mising mitchell sets we only had japan

MistyDreams:
pzloop2 and mpang added as per discussed in issue.


arcadez2003:
Thanks again

MistyDreams:
no probs

PR #1449, opened 2022-09-14
ddragon fix

MistyDreams:
this should be good save is on the issues page


mahoneyt944:
Hey the bot above is complaining about some declarations in this PR, can you verify they are good then I'll merge.

PR #1474, opened 2022-10-26
Hangon and system18 fill black

mahoneyt944:


mahoneyt944:
Yes it's enabled by default so I don't think it's an issue, so bombs away. https://github.com/libretro/mame2003-plus-libretro/blob/d5a70f9bf9601a3cc1773d5be6c3e83cbc287645/src/vidhrdw/system16_vidhrdw.c#L1067

PR #1481, opened 2022-11-16
Fix rom loading and add set 2 (woodpeca)

mahoneyt944:
use newer rom loading in pacman.c


mahoneyt944:
@arcadez2003 hey, I tried this game out and had to fix the rom loading for this core. Also added set  2. Does the mem map need ported too? Noticed it's slightly different than the pacman one.

PR #1496, opened 2022-12-16
add new resnet interface and fix dkong as an example

MistyDreams:
The res_net code was taken from mame0139.

@arcadez the legacy code is identical. I made sure it was with meld before splitting the header and c file. I added the normalize function as well as it was a little dull without it. I also had to add a total_color function for the normalization. I also left the fatal errors commented it in case you want to drop this on your xbox port.

This can potentially fix a few games with the new resnet code. Big thanks @timcannell for pointing out this needed fixed. 

the old red pallete init has also been added in case you need it its just been renamed to dkong_old

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -1043,8 +1043,8 @@ SOURCES_C += \
 	$(CORE_DIR)/machine/model1_machine.c $(CORE_DIR)/vidhrdw/model1_vidhrdw.c $(CORE_DIR)/drivers/model1.c \
 	$(CORE_DIR)/drivers/calorie.c \
 	$(CORE_DIR)/machine/mc8123.c \
-	$(CORE_DIR)/machine/namcoio_machine.c
-
+	$(CORE_DIR)/machine/namcoio_machine.c \
+	$(CORE_DIR)/vidhrdw/resnet.c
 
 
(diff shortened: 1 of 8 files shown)


mahoneyt944:
@MistyDreams this introduced some build issues
Screenshot_20221216-185124

mahoneyt944:
It's the declaration I'm palette.x line 1028. Declaration needs to be at the top of the for loop. Might be another let me look real quick

mahoneyt944:
Trying https://github.com/libretro/mame2003-plus-libretro/commit/71cb57025c677a1979abef3a032251175850bb48

mahoneyt944:
Good to go 👍

mahoneyt944:
Haven't actually looking into it. I just verify that all 28 (is it now?) platforms successfully build. When they don't, I look at their logs.

mahoneyt944:
Here's the ones that were failing though

Screenshot_20221216-190023

arcadez2003:
> What c standard are we compiling too? It might be worth setting it in the makefile to avoid any potential pitfalls.

Just plain ole C i think, some later code will be C++ as per MAME switching to that standard some time back so it 
requires slight modding back to C, anyway looks like it's all sorted now and good job BTW.

PR #1500, opened 2022-12-19
Nbajamte is parent

mahoneyt944:


mahoneyt944:
@MistyDreams do you agree with this? Later mame has this as its own parent

mahoneyt944:
It does share a few with nbajam, but I believe it's different enough that later mame separated them

arcadez2003:
No complaints from me

mahoneyt944:
sounds good. found this too which confirms train of thought
https://github.com/libretro/mame2015-libretro/commit/dad99e71075c0725940704aa040293bb37ab77ab

PR #1509, opened 2022-12-30
fix radarscope [mamedev] latest pallete and video update

MistyDreams:
Had a free hour decided to fix the rather nasty game. The screechy sound gets to me. The only real challenge was fixing the grid offset without changing the loop. The is two options to draw a timer and a function call. I left it as a function call as its less overhead. If this brings the game to be too slow just revert it I can fix it with a static background. I dont have access to a pi 3 to test.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -40,6 +40,12 @@ struct mame_bitmap
 	void (*plot_box)(struct mame_bitmap *bitmap,int x,int y,int width,int height,pen_t pen);
 };
 
+#define BITMAP_ADDR(bitmap, type, y, x)	\
+	((type *)(bitmap)->base + (y) * (bitmap)->rowpixels + (x))
+
+#define BITMAP_ADDR8(bitmap, y, x)	BITMAP_ADDR(bitmap, UINT8, y, x)
+#define BITMAP_ADDR16(bitmap, y, x)	BITMAP_ADDR(bitmap, UINT16, y, x)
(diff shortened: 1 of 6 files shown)


arcadez2003:
@MistyDreams You might find this commit of some interest https://github.com/finalburnneo/FBNeo/commit/67cd63d57549018af8052e3424e73196d0299665

arcadez2003:
Yeah i was a little rough myself yesterday all fine and dandy today, happy new year guys let's hope it's a good one as quite
honestly 2022 is worth forgetting, as for that commit @MistyDreams i think it's still preliminary ATM i just pointed it in your
direction as i remember you were working on improving the sound for Radar Scope and thought it might be of some 
interest.

Regards

arcadez2003:
TBH all that segfault data is gobbledegook to me you could always just send me the driver that is if you've not already
sussed it out.

arcadez2003:
> the samples should be hooked up like dkong all you really need is samples. MDRV_SOUND_ADD(SAMPLES, dkong_samples_interface) should play random dkong samples in theory at least.

Aye i know that i had the impression that these were gonna be wholly new samples created for Radar Scope.??  
maybe i got the wrong end of the stick, anyway FBN supports samples for certain games that MAME doesn't
this reminds me to look into porting the code for those across to here.

mahoneyt944:
@MistyDreams good work here. I'll merge when you're ready

PR #1591, opened 2023-05-25
Add a core option 'analog_sensitivity'.

wn2000:
The OSD menu allows changing analog sensitivity up to 255%. Sometimes this is not quite enough (tested on Arkanoid). So adding a core option with a larger range. The "default" value means to use the OSD menu setting. Any non-default value overrides the OSD menu. A core option is also easier to access than the OSD menu.

### DIFF
--- a/src/inptport.c
+++ b/src/inptport.c
@@ -873,7 +873,7 @@ void update_analog_port(int port)
 	}
 
 
-	sensitivity = IP_GET_SENSITIVITY(in);
+	sensitivity = options.analog_sensitivity ? (int)(options.analog_sensitivity * 100) : IP_GET_SENSITIVITY(in);
 	min = IP_GET_MIN(in);
 	max = IP_GET_MAX(in);
 	default_value = in->default_value * 100 / sensitivity;
@@ -1074,7 +1074,7 @@ static void scale_analog_port(int port)
 
 profiler_mark(PROFILER_INPUT);
 	in = input_analog[port];
-	sensitivity = IP_GET_SENSITIVITY(in);
+	sensitivity = options.analog_sensitivity ? (int)(options.analog_sensitivity * 100) : IP_GET_SENSITIVITY(in);
 
 	/* apply scaling fairly in both positive and negative directions */
 	delta = input_analog_current_value[port] - input_analog_previous_value[port];
@@ -1329,7 +1329,7 @@ profiler_mark(PROFILER_INPUT);
 						input_analog_init[port] = 0;
 						input_analog_scale[port] = 1;
 						input_analog_current_value[port] = input_analog_previous_value[port]
-							= in->default_value * 100 / IP_GET_SENSITIVITY(in);
+							= in->default_value * 100 / (options.analog_sensitivity ? (int)(options.analog_sensitivity * 100) : IP_GET_SENSITIVITY(in));
 					}
 				}
 				else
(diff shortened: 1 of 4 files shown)


mahoneyt944:
This isn't a good idea because each game uses a unique sensitivity, so a global setting here would cause issues between different games. Is there a certain game that's not working for you that you're trying to correct with this? 1000% sensitivity is a bit extreme, I can't imagine a practical use for that??

mahoneyt944:
Are you using an actual spinner, analog stick or dpad for Arkanoid?

wn2000:
Using an actual spinner (atgames legends ultimate cab).

mahoneyt944:
I bought mine from GRS Amazon store "GRS USB Button Hole Spinner".

mahoneyt944:
I just can't imagine needing to adjust it that high and set globally as you have it is not reasonable for most users, I use it at the stock settings and it plays fine, implementing this could cause more harm than good for users that aren't familiar with how mame is handling it. I would recommend adding this to a personal core, and just use it on the few games you're having issues with.

Ideally I would invest in a better spinner or see what other people with your cab are doing for solutions.

mahoneyt944:
Check this out

https://www.reddit.com/r/LegendsUltimate/comments/uucxv0/try_to_add_tsticks_spinner_to_legends_gamer_pro/

wn2000:
> I just can't imagine needing to adjust it that high and set globally

I would not set it globally. As a core option can be set either at the core / global level, or at the game level, the intention here is to only set it for the few games at the game level.

> I would recommend adding this to a personal core, and just use it on the few games

Yeah that's what I'm doing. Just thought it might be generic enough to be useful for other folks. But if not that's fine.

Before we close this though, I want to understand what is the real concern. Is it because

1. The OSD menu already provides a similar functionality so no need to have a core option for that, or
2. The OSD menu setting only applies at the game level, while the core option can be accidentally set at the core level, or
3. The values provided for that core option are too high.

It sounds like reason 2 to me. But RA settings are never "fool-proof" anyway. So I'd prefer "being able to do things" over "less possibility of making mistakes" (I wouldn't even say it's feature vs usability, as having that core option arguably improves usability than having to go into the osd menu).

mahoneyt944:
All 3 of those are reasons why. It also adds confusion since RA already supports analog sensitivity natively, which is different from what you're trying to do but the average person wouldn't know that. If a users enabled that option it will break all other games using sensitivity. Which is more issue tickets for us here.

Also seems to me you're doing unneeded calculations. you're dividing your var.value by 100 in core options then in the code multiply it by 100 again?

 It's adding a core option that shouldn't be needed with proper hardware. You should never need a range higher than what the OSD menu offers..... Not to mention that a game cab like you have directly breaks our licensing here. At games legend should fix it's issue with spinners so users like yourself don't have these issues in the first place.

That said the spinner i mentioned from GRS works well, it's what I use with this core, and specifically for Arkanoid as well as other games like tron.

wn2000:
> you're dividing your var.value by 100 in core options then in the code multiply it by 100 again?

Good point. May as well use the percentile as the internal representation.

> It's adding a core option that shouldn't be needed with proper hardware.

Although I still can't say I agree with the reasoning here, I do see that you have to take care of a larger user base than what I have in mind, and it's you folks providing direct support here. So closing for now. Appreciate you looking into this!

PR #1608, opened 2023-07-08
backport mamedev fixes for stlforce, add mortalr

MistyDreams:
@arcadez2003 I left the old sprite draw in just in case there is speed issues you can just change to the quicker sprites.  Stlforce will need a few tweaks the priorities are wrong in the old code but can get it looking the same as it did with the old code and both games are playable enough just a few niggles with priorities with the older code.

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -5592,6 +5592,7 @@ Other Sun games
 	DRIVER( sprcros2 )	/* (c) 1986 GM Shoji */
 	DRIVER( mugsmash )	/* (c) Electronic Devices (Italy) / 3D Games (England) */
 	DRIVER( stlforce )	/* (c) 1994 Electronic Devices (Italy) / Ecogames S.L. (Spain) */
+	DRIVER( mortalr )	
 	DRIVER( fantland )	/* (c) 1987 Electronic Devices Italy */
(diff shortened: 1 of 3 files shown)


arcadez2003:
@MistyDreams check sprites_commands_w a little compile issue with the int calls....

You may prefer it cleaner than this though :)

WRITE16_HANDLER(sprites_commands_w)
{
	int i;
	// TODO: I'm not convinced by this, from mwarr.cpp driver
	if (which)
	{
		switch (data & 0xf)
		{
		case 0:
			/* clear sprites on screen */
			for (i = 0; i < 0x400; i++)
			{
				sprites_buffer[i] = 0;
			}
			which = 0;
			break;

		default:
			// allow everything else to fall through, other games are writing other values
		case 0xf: // mwarr
			/* refresh sprites on screen */
			for (i = 0; i < 0x400; i++)
			{
				sprites_buffer[i] = stlforce_spriteram[i];
			}
			break;

		case 0xd:
			/* keep sprites on screen */
			break;
		}
	}
	which ^= 1;
}

arcadez2003:
> never hadnany compile issues what the problem

Ya must be lucky then i've always had that issue i think it's related to MAME switching to C++ rather than plain ole C 
but i could be wrong anyway we're good to go here then i'll look at the sound balances next week maybe unless someone
beats me too it.

PR #1609, opened 2023-07-08
compile fix

MistyDreams:
@mahoneyt944 though we where relaxed on the c99 stuff are // comments still ok?


arcadez2003:
Nah everything needs to be /**/ as per i got into trouble when i updated tecmosys.c a while back

arcadez2003:
Alright im getting rid of all the // i'll send the pr's shortly

arcadez2003:
Well either way it's all good now no harm done :)

mahoneyt944:
After removing the c99 restrictions, // comments work but out of habit I try to use /**/ as much as possible. Technically both are fine but it's safe habit to adopt to use /**/

PR #1615, opened 2023-08-08
update cps2 romsest to current mame since i fixed the decryption a while back

MistyDreams:
@arcadez2003 no source changes apart from the inputs and video configs that needed added. For you port all you need to do is copy the game definitions and the roms as well as the input and the video configs added.

@mahoneyt944 the 6 player games dont have the sf2 labels added so @arcadez2003  can copy this over to his ports if he wants too, Im not sure if every 6p game follows this naming convention either but will add them as that just making @arcadez2003 life a bit easier. 

This small update bags us all cps2 games the bootlegs need looked at. Some games may need tweaked in the gfx config. The video code is pretty old and might require updates time will tell as we never had all the sets

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -2484,6 +2484,7 @@ V-V                           TP-027
 	DRIVER( progear )    /* Cave (Capcom license), Progear (USA 010117) */
 	DRIVER( progearj )   /* Cave (Capcom license), Progear no Arashi (Japan 010117) */
 	DRIVER( progeara )   /* Cave (Capcom license), Progear (Asia 010117) */
+	#if defined(CPS2_BOOTLEG)
 	DRIVER( ddtodd )     /* bootleg, Dungeons & Dragons: Tower of Doom (Euro 940412 Phoenix Edition) (bootleg) */
 	DRIVER( ecofghtrd )  /* bootleg, Eco Fighters (World 931203 Phoenix Edition) (bootleg) */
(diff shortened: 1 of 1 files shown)


mahoneyt944:
Are we still using the bootstraps?

mahoneyt944:
It was requested at one point, I think for max players and the coin chute configs in the service menu. Been awhile since I looked at it though.

arcadez2003:
@MistyDreams no need to bother working the code around my Xbox ports as i no longer bother with em but thanks
for the thought :)

mahoneyt944:
Might be problematic, the bootstraps were region specific if I recall, so idk if they will just work for all the new sets.

mahoneyt944:
Don't remember how I set ithe chutes before tbh. You'd have to reference the core pre commit or make it however you think is best. Pretty sure the max players was the main issue and I made a judgement call on the chutes

markwkidd:
really nice!

arcadez2003:
> > @MistyDreams no need to bother working the code around my Xbox ports as i no longer bother with em but thanks
> > for the thought :)
> 
> I was under the impression seems wrongly that you ported to a few different cores. If the changes just make it to this core its a little less motivating. Never the less its another one for the pot.

Well i used too it's just i dont have the time anymore TBH.

mahoneyt944:
Speaking of, just say when you want this pushed. Wasn't sure if you're still working or not but will merge when you say it's ready.

PR #1622, opened 2023-09-09
fix logerror and change pgm printf to logerror

MistyDreams:
Some other drivers are using this but it wasnt implemented correctly. Also changed the pgm  printf to logerror. logerror redirects to retroarch debug log.


arcadez2003:
Thanks as always

PR #1629, opened 2023-09-21
Spriteram is now shared

arcadez2003:


arcadez2003:
@mahoneyt944 as per this commit https://github.com/mamedev/mame/commit/064401bf006252f8d87f01ead3e9aa86e45c3433 

It looks like the spriteram is now shared rather than writeonly  so would require for us to have these calls on the read map.??

mahoneyt944:
Can't see any harm in it.

PR #1643, opened 2023-10-20
fix Fantastic Journey

grant2258:
@arcadez2003 here is the fix for you issue hope you don't mind as you done the legwork. The controls need looked at. Ill have a peek when I have more time but this will sort black screen


grant2258:
Inputs are fine was my default config that was bad so consider it working now.

PR #1662, opened 2024-01-11
model1 running as it should log was spamming the function slowing thi…

grant2258:
Disable processing of logerror since the model1 is spamming it and it getting processed even if its not printing severely slows down model1 games with the amount of spamming sent. Just jest this define if you need error logs only devs would need access to this anyway.


grant2258:
@KMFDManic for the original port retropie branch has the fsigned fixes here https://github.com/RetroPie/daphne-emu/commit/f35231da6dc2238d4dedb74e18f528f2e5919284. The libretro makefile will need these flags added.

```
CFLAGS +=-fsigned-char
CXXFLAGS +=-fsigned-char
```

add this at line 13 and you should be good to go.

PR #1670, opened 2024-01-21
add aquario no sound atm (arcadez wip) 

grant2258:
The sound still needs fixed other wise its working

Screenshot from 2024-01-21 18-26-41

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -6097,5 +6097,5 @@ Other Sun games
 	DRIVER( roundup5 )
 	DRIVER( bigfight )
 	DRIVER( cyclwarr )
-
+DRIVER( aquario )
 #endif	/* DRIVER_RECURSIVE */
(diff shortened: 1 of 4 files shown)


grant2258:
@arcadez2003 this seems to affe
Screenshot from 2024-01-21 18-36-14
ct the shadow in moon walker for some reason

mahoneyt944:
Probably the machine driver change, try making an isolated machine driver for vdp only

grant2258:
doubtful its the machine driver change only change there was the pallet length. Most likely the vdp changes. Ill leave it as is arcadez can take a look ill revisit this another way at the weekend if no progress is made :).

grant2258:
@arcadez2003 thats it done the code in this core driver is very picky about the palette size being changed I never seen and errors either after changing it . The sounds fixed moonwalker is fine. This is probably as good as your going to get with the old code unless someone updates it but its perfectly playable if you can live with a few gfx glitches.

arcadez2003:
Ok Let's merge it then the only games affected here are system18 and segac2 due the video changes
we can always look out for any bug reports concerning those type games down the road.

PR #1673, opened 2024-01-24
use banking for braze

grant2258:
@arcadez2003 managed to lookup the banking code in mame. Ive converted our driver to banking dongx is ok but dkremix isint even making a bank request im thinking it might be failing to boot due to dma. dkongx tells us this is happening but you can get past it by pressing a key.

 
Screenshot from 2024-01-24 21-14-12

can close this if you dont wan tthe banking changes as it doesnt fix the dkremix issue.


grant2258:
@mahoneyt944 added xmas remix that one is working fine.

PR #1685, opened 2024-02-10
Hamaway

arcadez2003:


grant2258:
@mahoneyt944  just change it to return 0xff;. If you can enable ld flags to show unresolved symbols at linking time for the future that would be helpful in the long run.

mahoneyt944:
Should be able to do that in the make file, I don't have access to the actual buildbot though.

grant2258:
make file is where to do it if you want to fix it is up you will cause less headaches in the long run.

PR #1705, opened 2024-04-07
add toprollr as per request

grant2258:
Latest mame had a different crc from the old base looks the same as mame current so removed the flag! No need to vredit me all I did was the banking :). You might want to let the fbneo peeps this game isint added if they want to port it. 
Screenshot from 2024-04-07 19-42-39

### DIFF
--- a/src/driver.c
+++ b/src/driver.c
@@ -320,6 +320,7 @@ const struct GameDriver *test_drivers[] =
 	DRIVER( silvland )	/* Falcon */
 	DRIVER( yamato )	/* (c) 1983 Sega */
 	DRIVER( yamato2 )	/* (c) 1983 Sega */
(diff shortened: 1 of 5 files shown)


arcadez2003:
Well without the banking we'd not be rolling so you gotta be credited bigman, as for FBN i mentioned this game
to dink last year and he thought it worth adding i suppose he's not got around to it yet.

grant2258:
well was worth mentioning noticed it wasnt there when I was looking for inspiration but I see how mame was handling the memory and just done a memcpy of the bank and opcode base :)

arcadez2003:
Need to ask ya with regards to the driver.c about hamaway being removed unless it's just your file was slightly different
than ours and wee have that one declared elsewhere in the romlist

grant2258:
it was the files you gave me in the wip

PR #1732, opened 2024-05-20
gotFrame main run loop gain

mahoneyt944:


mahoneyt944:
Copied this over manually, look good?

mahoneyt944:
@grant2258 Copied this over manually, look good?

grant2258:
No Its not ready to go something different on this core.

mahoneyt944:
Seems like the API might still expect video_cb to send that Null buffer? But I'm not sure. I'd have to check.

grant2258:
its not that I think I have it it was foo bared. Setting gotFrame to one. Just need to do a few more tests. We where rendering more than we needed too.

PR #1737, opened 2024-05-22
fix frameskip take two

grant2258:
The variables need set in retro run and there was an if statement I removed but I didnt do a git pull. the frameskip call is requested at least two times a frame so its set in retro run now(some drivers call it as well. The manual frame skips working a lot better now. Ill leave any updates to weekend when I have time to test properly in future.

dont pull this yet I want to test some more things. Ill drop you a wee note here when im done.

### DIFF
--- a/src/mame2003/mame2003.c
+++ b/src/mame2003/mame2003.c
@@ -388,6 +388,8 @@ void cpu_pause(bool pause)
     cpu_pause_state = false;
 }
 
+extern UINT8 frameskip_counter;
+
 void retro_run (void)
 {
   bool updated = false;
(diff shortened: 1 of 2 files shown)


grant2258:
ok we are good to go aggressive max drops a very few frames. I looked at the low latency there is no point setting it unless you want to increase it it wont let you go lower than the setting in the front end.

grant2258:
@mahoneyt944 fixed the interupt pause issue as well.

mahoneyt944:
> @mahoneyt944 fixed the interupt pause issue as well.

This fixed any issues I found with the interrupt https://github.com/libretro/mame2003-plus-libretro/commit/5f0dff30123f6f891f15a577900ebdc17799d775 

Did you find something that failed with that commit still? The driver is still running when paused so event_index was indexing while paused.

grant2258:
yes midres fails checked it while i was testing the frameskip on a few games.

mahoneyt944:
@grant2258 do me a favor and submit the pause fix in a separate PR, I want to bench test it with a boat load of games.

grant2258:
separated the two is quicker

mahoneyt944:
I copied the changes into a branch, it fixed midres but broke lethalen. So for now just submit the frameskip bits and I'll keep working on the pause. Gonna try something.

mahoneyt944:
If you want dump the frame skip update in mame2003 too. I did the part 1 already

grant2258:
I need to get to bed for work gluck.

PR #1778, opened 2024-07-07
Soundcore fix

grant2258:
@mahoneyt944 thats it fixed and hamaways missing sound as well technically you dont even need this update just fix the data range on the core update you done earlier.
It will have bugs compared to this one though. I was a ten minute fix to be honest so not really caring what you do with it.

### DIFF
--- a/src/drivers/segas32.c
+++ b/src/drivers/segas32.c
@@ -2856,7 +2856,7 @@ INPUT_PORTS_END
 struct RF5C68interface sys32_rf5c68_interface =
 {
   12500000,	/* pitch matches real PCB, but this is a weird frequency */
-  55
+  100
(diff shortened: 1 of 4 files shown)


mahoneyt944:
Do we need the clock change on system18 too or no?

mahoneyt944:
I only see the interface in `src/machine/system16_machine.c` is there another hiding? Or they all using that one?

grant2258:
Well to be honest you would need to check with mame current if they are all using the same they are on this core though

mahoneyt944:
Hmm seems like it's only connected to sys18 and not system16 so maybe it's not that much of an issue. Just use the sys18 clock I guess.

PR #1790, opened 2024-07-16
update

grant2258:
@mahoneyt944 try this you missed a few updates towards the indirect palette fix. Some of the gfx rom load offsets changed check the on  mame0104u7. No real work needed dont the code just copied over fine.  See if that gets you fixed up.


mahoneyt944:
You have to lay current mame over our video driver, this commit is removing the fixes after 2015.

The issue with sonic has been there through all the revisions of the video code I 
tried (and on sonicp as well), which makes me think it might be in the machine driver or related to the trackball input maybe?

arcadez2003:
@mahoneyt944 Have you sorted the crash on the ice level.?? if not i might have a lead towards a fix that was done
MAME99 with no mention on the changelogs.

grant2258:
@mahoneyt944 can you point to the commit your talking about from 2015. This makes little sense as the new offsets are used in current mame.

mahoneyt944:
> @mahoneyt944 Have you sorted the crash on the ice level.?? if not i might have a lead towards a fix that was done by MAME99 with no mention on the changelogs.
> 
> I mean some of these guys are muppets but they defo mention a fix which i assume to be between 98-99 but who knows what versions they used before 99, anyway if you've fixed this then feel free to ignore :) https://forums.sonicretro.org/index.php?threads/mame-now-fully-supports-segasonic-arcade.5140/

Yes that was caused by the video code trying to draw outside the bitmap space. 

@grant2258 not sure what offsets your talking about. The commits though are spread out over a few.

Backdrop color
https://github.com/mamedev/mame/commit/90152f33d5114828a53f38fc9a1913af71e683d6

Color wrap
https://github.com/mamedev/mame/commit/0f5b710276ab456d4c5f8027c9b9ca8c1ccdaae5

Global flip / compute flip
https://github.com/mamedev/mame/commit/3ca05f596d14fba033f33fc4ec1dbb28d8617070
https://github.com/mamedev/mame/commit/13b609deb2076a9c5df4e55dc042e1fc3c6f3988
https://github.com/mamedev/mame/commit/770d93fa483cc78acfbce563d12602664a28a166

And some references to current mame misc changes
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/mame/sega/segas32_v.cpp#L402
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/mame/sega/segas32_v.cpp#L539
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/mame/sega/segas32_v.cpp#L1031
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/mame/sega/segas32_v.cpp#L1510
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/mame/sega/segas32_v.cpp#L1451


One change we might still need is use of this utility function `sext`
https://github.com/mamedev/mame/blob/3aa9c6a1f8fb749bbfaf431c04afa92cd0b09d2b/src/lib/util/coretmpl.h#L671

mahoneyt944:
@grant2258 I also just did a test with your commit here and the issue is still present. The game is just hard to control though. Definitely going to eat quarters. If you're up for trying to fix the savestate I can give you a good save state to use. Seems like the issue is either outside the video code or a fundamental issue with the core. I would like to try to hookup they `sext` until function though to see if that helps or not.
 
The save state fix is part of this commit
https://github.com/mamedev/mame/commit/90152f33d5114828a53f38fc9a1913af71e683d6

grant2258:
@mahoneyt944 the main change to test in there was the sprite banking changes. I can see a few backports where removed but the sprite banking changes where missed

mahoneyt944:
> If you load dbzvrvs youll see

Yeah that's a known issue with the current video code
https://github.com/libretro/mame2003-plus-libretro/blob/1ae2425549aa92a613123f3fa4085f9b3f386487/src/vidhrdw/segas32.c#L38-L39

Last night I did a full playthrough of every game in the driver to try to find issues unique to mame2003+ but not current mame since we are echoing their video code. Sonic and sonicp are the only ones I found that had a weird issue which makes me believe the solution might be outside the video code. Maybe input related or something.

grant2258:
Here is an example with the romload/bank fix f1en wont work properly on the old code with the propper romloads the way the banks are calculated. It worth keeping in mind when adding other roms of the banking changes even if it doesnt fix the issue at hand directly. Ill try my test core if i can find cheats for sonic.

grant2258:
can close this for now @mahoneyt944 ill test it a  mame0104u7 updates on the test core.

mahoneyt944:
> Here is an example with the romload/bank fix f1en wont work properly on the old code with the propper romloads the way the banks are calculated. It worth keeping in mind when adding other roms of the banking changes even if it doesnt fix the issue at hand directly. Ill try my test core if i can find cheats for sonic.

Oh that's interesting, didn't realize some of them were simplified down. I'll look into those changes soon.

grant2258:
sonic seems to work fine on the mame0104u7 updates on the test core. Ill test more a little later on at some point. I did have odd things happening with thunder blade when I hooked up the interrupts properly. The old inputs would just not work properly was fine when I updated the  inputs. That would be a good place to start from confirm the interrupts.

PR #1802, opened 2024-07-21
update segaxbd

grant2258:
Updated interrupts and io to mame095 smgp still isint working get credit sounds and a bit of a statuc picture and no unmapped read/writes. I also dumbed down the logging for sonic and this driver will only dumb it down as needed can be enabled buy setting verbose like the original code was. This will need to be looked at some other time.

### DIFF
--- a/src/cpu/z80/z80.c
+++ b/src/cpu/z80/z80.c
@@ -85,6 +85,14 @@
 #include "z80.h"
 #include "log.h"
 
+#define VERBOSE 0
+
+#if VERBOSE
+#define LOG(x)	logerror x
+#else
+#define LOG(x)
+#endif
+
 /* execute main opcodes inside a big switch statement */
 #ifndef BIG_SWITCH
 #define BIG_SWITCH			1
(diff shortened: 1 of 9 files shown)


arcadez2003:
smgp works without the extra comms code on my xbox driver it still has the CPU4 in the rom table but it seems not to matter
so if your mem mapping is fine for it maybe it's the divider code in the machine.?? or maybe you need to check your using the
correct decrypted roms for the set your using as there are many versions of that game supported.

Anyway i added some stuff needed by Line Of Fire dunno if it'll work though we'll see.

grant2258:
@arcadez cant use the sets in earlier mames as they are encrypted unless we add decryption. It seems something to do with the sound cpu. I was going to get round to loffire  but been messing with dmgp we get a static screen now going to add the nvram handler might  just provide a mircle

arcadez2003:
> @Arcadez cant use the sets in earlier mames as they are encrypted unless we add decryption. It seems something to do with the sound cpu. I was going to get round to loffire but been messing with dmgp we get a static screen now going to add the nvram handler might just provide a mircle

Aye i know that bigman it's just i saw you added the comms and were talking about needing the 4th CPU so i thought
it worth posting that my driver doesn't have that code nor that the game seems to need it anyway it seems your
making progress and knowing how ya have a knack for this im sure everything will be good soon enough :)

grant2258:
yea i notice it wasnt hooked up to anything but added it the sake of it. Strange thing the game crashes with the init. I Ill try move it to machine init is worth a try.

PR #1921, opened 2025-01-21
fix 64street attract

grant2258:

### DIFF
--- a/src/drivers/megasys1.c
+++ b/src/drivers/megasys1.c
@@ -369,11 +369,10 @@ static MEMORY_READ16_START( readmem_C )
     { 0x0f0000, 0x0f3fff, MRA16_RAM },
     { 0x0f8000, 0x0f87ff, paletteram16_word_r },
     { 0x0d8000, 0x0d8001, ip_select_r },
-    { 0x0f0000, 0x0f3fff, MRA16_RAM },
-	{ 0x1c0000, 0x1cffff, ms1_ram_r },
-	{ 0x1d0000, 0x1dffff, ms1_ram_r },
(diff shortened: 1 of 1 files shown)


grant2258:
all games are working fine accept chimerab. I even added your maps in because I was messing about with it. Its working fine on my port. If you dont get any progress with it. We can add the old memory reads back for it.

mahoneyt944:
Ok, I'll push this and see if I can get that one going with the change.

mahoneyt944:
I think the issue is with the read, `ms1_ram_r` probably needs to check LSB and MSB like the write.

grant2258:
Ill put a pull request in at some point with the old memory map. It could be the the memory map just needs fixed but would rather have it working than not. We can come back to review it at some point.

mahoneyt944:
> Ill put a pull request in at some point with the old memory map. It could be the the memory map just needs fixed but would rather have it working than not. We can come back to review it at some point.

I went ahead and pushed a commit for this, hopefully we can figure out why it's not playing nice. If its working in your core, might be some memory difference, idk.

grant2258:
I was hoping to get the chance to look at it last night. The only think I can think of is its setting the opcode base for the cpu in one of these regions memory regions. From the logs  MEMORY_ADDRESS_BITS(21) changing  0x1f0000, 0x1fffff, ms1_ram_w read and writes should cause the game to boot. It wont work properly though. I just need to think a solution round it if there are any.

mahoneyt944:
Yeah it's gets an invalid instruction for some reason. Maybe it needs something init to prevent that.

PR #1922, opened 2025-01-23
cyclone: use HAVE_ARMv6 flag for backward asm compatibility

Apaczer:
after 5f98dc3cf18cee1e2e9d0264d5851f612642d6e7 the build fails for old armv5 platform when using CYCLONE, so for compatibility use HAVE_ARMv6 in preprocessed cyclone.S (e.g. use in [upstream](https://github.com/notaz/cyclone68000/blob/3ac7cf1bdeecb60e2414980e8dc72ff092f69769/OpAny.cpp#L197))

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DEBUG         ?= 0
 DEBUGGER      ?= 0
 SPLIT_UP_LINK ?= 0
 ARM           ?= 0 # set to 0 or 1 to indicate ARM or not
+HAVE_ARMv6    ?= 0
 CPU_ARCH      ?= 0 # as of November 2018 this flag doesn't seem to be used but is being set to either arm or arm64 for some platforms
 
 LIBS          ?=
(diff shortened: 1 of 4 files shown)


grant2258:
You essentially changing all platforms that dont need this update to be slower. Im not going to be looking over this stuff after adding it and getting it working. Make sure your using the cyclone from  https://github.com/notaz/cyclone68000 and not something older as the source for the 68k if your wanting these changes. Im sure someone will come along and say hey this isint ideal for pi2 pi3 pi4 and mini classics. Since im not using it anyway im not too bothered.

mahoneyt944:
@grant2258 so you think the added arm flag will slow down platforms? If so is there an alternative solution?

Apaczer:
I believe only miyoo_arm32 is using pre ARMv6 instructions, so it would be valid to define HAVE_ARMv6=1 in one place for all platforms with exception to that one. 
Roughly speaking it was this (add C macros in the same ASM code), or split them into two files to follow current coding style. Would be the best to generate cyclone code at buildtime by fetching submodule from notaz repo and pass correct flags for specific platform like for e.g in picodrive

grant2258:
The problem is you need to define HAVE_ARMv6 = 1 in your platform you only defined arm in the other commit see https://github.com/libretro/mame2003-plus-libretro/blob/a5bb1f44251aaad1abd36bd7a56159fc03b88a16/Makefile#L857-L859

grant2258:
ah I thought the miyoo was more recent hardware my bad.

PR #1927, opened 2025-01-26
Align map, clean, missing reads

mahoneyt944:


mahoneyt944:
@grant2258 hey, I noticed the addresses were out of order and missing some reads. Not sure if these reads were just missed previously or left out intentionally. Figured I'd get a second opinion before submitting this clean up PR to be sure.

grant2258:
Shouldn't do any harm I was looking at mapped memory on the different games and added as needed for debugging. The maps we inherited where out of order. I didn't realise that was an issue on issue this core or spacing its literally littered with code. If it does cause issues we can check what's going wrong.

PR #1930, opened 2025-01-28
update pengo sets and add missing

grant2258:
Played this game as a nipper but the the version the local arcade had wasnt popcorn music


arcadez2003:
Always good to have your pick of the pengo's :)

grant2258:
Honesty added all sets to try recreate that issue thats going on cant manage it. Pretty sure its something there end but need to be thorough cant fix something I cant make happen.

PR #1949, opened 2025-03-18
Makefile: platform=miyoo ->USE_CYCLONE & USE_DRZ80

Apaczer:
add necessary ARM assembly opt. for Miyoo ARMv5 to make core even work.

In our downstream fork we are using also following setup:
```diff
diff --git a/src/mame2003/core_options.c b/src/mame2003/core_options.c
index 2e35ba1a..1ce76792 100644
--- a/src/mame2003/core_options.c
+++ b/src/mame2003/core_options.c
@@ -768,7 +768,11 @@ static struct retro_core_option_v2_definition option_def_cyclone_mode = {
       { "Cyclone+DrZ80(snd)", NULL },
       { NULL, NULL },
    },
+#if defined(LOW_ARM)
+   "Cyclone+DrZ80(snd)"
+#else
    "default"
+#endif
 };

 static struct retro_core_option_v2_definition option_def_override_ad_stick = {

```

Otherwise core hangs with "default" cyclone_mode, but I'm not sure if maintainers would be ok with such change?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -532,11 +532,14 @@ else ifeq ($(platform), miyoo)
 	CC = /opt/miyoo/usr/bin/arm-linux-gcc
 	CXX = /opt/miyoo/usr/bin/arm-linux-g++
 	AR = /opt/miyoo/usr/bin/arm-linux-ar
-	fpic := -fPIC
+	fpic := -fno-PIC
 	LDFLAGS += -shared -Wl,--version-script=link.T -Wl,-no-undefined
 	PLATCFLAGS := -DNO_UNALIGNED_ACCESS
 	PLATCFLAGS += -fomit-frame-pointer -march=armv5te -mtune=arm926ej-s -ffast-math
 	CXXFLAGS += -fno-rtti -fno-exceptions
+	ARM = 1
+	USE_CYCLONE := 1
+	USE_DRZ80 := 1
 
 # Emscripten
 else ifeq ($(platform), emscripten)


mahoneyt944:
Do you know why it hangs when set to default? I'd be more inclined to fix the default behavior then to bandaid it. I don't have this platform to test however.

grant2258:
Some games wont work with cyclone you need to change the defaults if that is the case. see https://github.com/libretro/mame2003-plus-libretro/blob/a5bb1f44251aaad1abd36bd7a56159fc03b88a16/src/mame2003/frontend_list.h#L9 for the list change the setting there.

grant2258:
@Apaczer can you make sure the option is set to `default` like it should and not `Cyclone+DrZ80(snd)`  and open the game called rygar. This has two z80s I need to see if its a specific game or a problem with the drz80 cpu core on your platform.  This is one of the games that does work with the drz80 core. I used during testing. 

If its just one specific game that is freezing you need to update the frontend_list with the game that is not working to the correct value for default. I have no idea who created this list but it could possibly need updating. We cant change the defaults to a specific game that could cause another game to freeze as a default option. We could disable it by default or use forntend_list.

Apaczer:
I believe it was because of core size and disabled Cyclone in build so it was loading indefinitely on weaker platform (bad word usage of "hangs"), also I saw smb inclining to switch "default" to smth else in issue section. Now with latest builds I could not replicate, sorry for fuzz (no need for patch @mahoneyt944).
Thanks @grant2258 for guides, I think with roms using `0=None` in frontend_list.h it will be unplayable on such low-end arm device, but either setting it now to OFF seems to be defaulting to Cyclone (at least assuming from performance lvl).

grant2258:
you can update the frontend_list some are set to zero because the games dont work with one or both the asm cores. If you think this list is wrong you can put a pull request i to fix it.

PR #1952, opened 2025-03-25
sound fixes

grant2258:
* fix missing segas and multi 32 channel and bring in the stereo like it should be
* fix rf5 core

### DIFF
--- a/src/ost_samples.c
+++ b/src/ost_samples.c
@@ -506,7 +506,7 @@ struct Samplesinterface ost_mk =
 struct Samplesinterface ost_moonwalker =
 {
 	2,	/* 2 channels*/
-	100 , /* volume*/
+	100, /* volume*/
 	moonwalker_sample_set_names
 };
 


grant2258:
@mahoneyt944 test any games you have had issues with on sega/multi32 with sound I tested with jpark bug you will notice a big difference with the ost as well as the mixer was shooting right up. It was always like that as you can hear from the xtreme samples. Just our samples got louder as well.

mahoneyt944:
I just look at the pipeline when it builds all the platforms (direct link at the top of the readme page). Im not sure if libretro would want to link the errors or not. I can ask.

grant2258:
Im not too bothered but the gcc flags would be handy to avoid this scenario ask them what flags we need to make gcc mimic it.

grant2258:
ok that should be it fixed would be ideal if you can possibly get them flags dont have enable them just note makefile if possible.

mahoneyt944:
Well apparently to get the output you requested, you'd have to just use an older gcc. I don't know that there is a way to get what you want through flags. Though if you figure out a way to do so, I'd add it for you... but couldn't figure a way to do it.

grant2258:
Ok pitty I thought there would be some easier way to fix fix these bugs than waiting on the build bot.

mahoneyt944:
C89 does not support Variable Length Arrays (VLAs). VLAs were introduced in C99. In C89, the size of an array must be a compile-time constant. Therefore, you cannot declare an array with a size determined by a variable at runtime and initialize it directly.

grant2258:
Who told you this @mahoneyt944 ?

mahoneyt944:
From my understanding, libretro strives to be compliant in this area, but the reality is most modern compilers handle this kind of stuff and the majority of people are using modern compile environments.... But they still at least aim for compliance.

grant2258:
The only compliance being aimed for is msvc failures there is certainly not any strict compliance to c89 and RA uses -std=gnu99 as the standard on unix when compiling. Its certainly entertaining when these issue get asked about. This core wouldn't compile neither would most with compliance enforced. Its a bit of a running joke to support old msvc by the looks of things.

mahoneyt944:
I don't have control over what they aim for or enforce. I just try to see that the cores we automate actually build. If you feel strongly about msvc or any other build practices, I'm sure the libretro forums or the RA repo itself might be able clear up some of this for you. It's quite possible that this core and many others are simply just not at the compliance level they would like them to be.

grant2258:
Well Retroarch itself doesn't comply with c89, it would be a complete waste of time talking into a tin can.  I do have a clearer picture now code must work on old MSVC compilers as a minimum rather than strict c89 compliance. There nothing to discuss on that front. I know how to apply strict compliance to fix any of them issues on newer compilers when msvc kicks and screams on the buildbot.

grant2258:
have added a pull request fix  in so clang and gcc will fail with this type or error if more come up we can add them. This way people locally will know there is an issue with the code. Its good for us that done use the buildbot to know there is an issue to fix. Doing something practical to fix these seems more sane than debating  interpretations and the only failures we see are msvc ones.

PR #1953, opened 2025-03-25
silence msvc

grant2258:
I dont really want to use malloc and free its expensive with cpu cycles so this will have to do


grant2258:
oh dear lord will need to figure out msvc defines and make that use malloc and free. Its doesnt seem to like dynamic sized arrays

arcadez2003:
> oh dear lord will need to figure out msvc defines and make that use malloc and free. Its doesnt seem to like dynamic sized arrays

Just when ya think your done they pull ya back in :)

grant2258:
MSVC should be abolished m8 :)

PR #1954, opened 2025-03-26
back in the time machine compiler support

grant2258:
Will use malloc and free from msvc only.


mahoneyt944:
Do you have any idea what value is typically set in `length`?

grant2258:
Its a movable target if frames are and depending what sample rate you choose. I know whats happening can be fixed other ways as well. Ill revisit this to make it not need the extra buffers. Was just a quick fix. It was always broke even in the old rf5.

PR #1956, opened 2025-03-27
Fix the rf5c68 another way

grant2258:
@mahoneyt944 Im not sure what you issue was wait the way it was. Regardless here is another fix for it.

how the ost sounds 
[mwalker.zip](https://github.com/user-attachments/files/19492368/mwalker.zip)

There is a problem in moon walker ost stops working if you reset the games not a game stopper just letting you know.

### DIFF
--- a/src/sound/rf5c68.c
+++ b/src/sound/rf5c68.c
@@ -39,14 +39,9 @@ struct rf5c68pcm *chip;
 
 static void rf5c68_update( int num, INT16 **buffer, int length )
 {
-	#if defined _MSC_VER
-	INT32 *left =  (INT32*)malloc((length) * sizeof(INT32));
-	INT32 *right = (INT32*)malloc((length) * sizeof(INT32));
-	#else
-	INT32 tempbuffer[2][length];
-	INT32 *left =  tempbuffer[0];
-	INT32 *right = tempbuffer[1];
-	#endif
+
+	INT16 *left =  buffer[0];
+	INT16 *right = buffer[1];
 	int i, j;
 
 	/* start with clean buffers */
@@ -90,13 +85,13 @@ static void rf5c68_update( int num, INT16 **buffer, int length )
 				if (sample & 0x80)
 				{
 					sample &= 0x7f;
-					left[j] += (sample * lv) >> 5;
-					right[j] += (sample * rv) >> 5;
+					left[j] += (sample * lv) >> 6;
+					right[j] += (sample * rv) >> 6;
 				}
 				else
 				{
-					left[j] -= (sample * lv) >> 5;
-					right[j] -= (sample * rv) >> 5;
+					left[j] -= (sample * lv) >> 6;
+					right[j] -= (sample * rv) >> 6;
 				}
 			}
 		}
@@ -105,21 +100,24 @@ static void rf5c68_update( int num, INT16 **buffer, int length )
 	/* now clamp and shift the result (output is only 10 bits) */
 	for (j = 0; j < length; j++)
 	{
+		UINT8 output_shift=10;
+		INT32 output_nandmask = (1 << output_shift) - 1;
 		INT32 temp;
+		
 		temp = left[j];
-		if (temp > 32767) temp = 32767;
+		temp *= 2;
+
+		if (temp > 32767 || temp > 32767 ) temp = 32767;
 		else if (temp < -32768) temp = -32768;
-		buffer[0][j] = temp & ~0x3f;
+		buffer[0][j] = temp & ~output_nandmask;
 
 		temp = right[j];
+		temp *= 2;
+
 		if (temp > 32767) temp = 32767;
 		else if (temp < -32768) temp = -32768;
-		buffer[1][j] = temp & ~0x3f;
+		buffer[1][j] = temp & ~output_nandmask;
 	}
-	#if defined _MSC_VER
-	free(left);
-	free(right);
-	#endif
 }
 
 
@@ -141,16 +139,16 @@ int RF5C68_sh_start( const struct MachineSound *msound )
 	int i;
 
 	if (Machine->sample_rate == 0) return 0;
-
+	
 	chip = auto_malloc(sizeof(*chip));
-
-	memset(chip, 0, sizeof(*chip));
+	
+	memset(chip, 0, sizeof(*chip));	
     /* f1en fix bad sound if set initialized to 0xff fixed in mame0215*/
 	for (i = 0; i < 0x10000; i++)
 		chip->data[i]=0xff;
 
-	//intf = inintf;
-
+	//intf = inintf;	
+	
 	name[0] = buf[0];
 	name[1] = buf[1];
 	sprintf( buf[0], "%s Left", sound_name(msound) );
@@ -160,7 +158,7 @@ int RF5C68_sh_start( const struct MachineSound *msound )
 
 	chip->stream = stream_init_multi( 2, name, vol,  inintf->clock / 384 , 0, rf5c68_update );
 	if(chip->stream == -1) return 1;
-
+	
 	return 0;
 }
 
--- a/src/sound/rf5c68.c
+++ b/src/sound/rf5c68.c
@@ -103,7 +103,7 @@ static void rf5c68_update( int num, INT16 **buffer, int length )
 		UINT8 output_shift=10;
 		INT32 output_nandmask = (1 << output_shift) - 1;
 		INT32 temp;
-		
+
 		temp = left[j];
 		temp *= 2;
 
@@ -139,16 +139,16 @@ int RF5C68_sh_start( const struct MachineSound *msound )
 	int i;
 
 	if (Machine->sample_rate == 0) return 0;
-	
+
 	chip = auto_malloc(sizeof(*chip));
-	
-	memset(chip, 0, sizeof(*chip));	
+
+	memset(chip, 0, sizeof(*chip));
     /* f1en fix bad sound if set initialized to 0xff fixed in mame0215*/
 	for (i = 0; i < 0x10000; i++)
 		chip->data[i]=0xff;
 
-	//intf = inintf;	
-	
+	//intf = inintf;
+
 	name[0] = buf[0];
 	name[1] = buf[1];
 	sprintf( buf[0], "%s Left", sound_name(msound) );
@@ -158,7 +158,7 @@ int RF5C68_sh_start( const struct MachineSound *msound )
 
 	chip->stream = stream_init_multi( 2, name, vol,  inintf->clock / 384 , 0, rf5c68_update );
 	if(chip->stream == -1) return 1;
-	
+
 	return 0;
 }
 


mahoneyt944:
That's because sample support breaks when you reset.

grant2258:
> That's because sample support in general breaks when you reset.

Thats a strange answer, ill check games that use samples to see if its the case like dkong ect. It certainly shouldn't stop working for everything after a reset. Will report back when ive checked this.

grant2258:
Good news I got it working but I need to do some testing on the samples with loading and unloading on the bigger samples. The original codebase had this issue with the samples after a reset as well.

grant2258:
I was working on updating the samples got bored with it just not something im motivated to do. Anyhow here is the fix for the samples not sure why this was left broke if if was know,

```
diff --git a/src/common.c b/src/common.c
index 3931b6dd..f6157f6a 100644
--- a/src/common.c
+++ b/src/common.c
@@ -322,7 +322,7 @@ static struct GameSample *read_wav_sample(mame_file *f, const char *gamename, co
 
 		/* allocate the game sample */
 		if(b_data == 1)
-			result = auto_malloc(sizeof(struct GameSample) + length);
+			result = malloc(sizeof(struct GameSample) + length);
 		else
 			result = malloc(sizeof(struct GameSample));
 
@@ -422,7 +422,7 @@ static struct GameSample *read_wav_sample(mame_file *f, const char *gamename, co
 		}
 
 		if (b_data == 1)
-			result = auto_malloc(sizeof(struct GameSample) + (flac_file.total_samples * (flac_file.bits_per_sample / 8)));
+			result = malloc(sizeof(struct GameSample) + (flac_file.total_samples * (flac_file.bits_per_sample / 8)));
 		else
 			result = malloc(sizeof(struct GameSample));
 
@@ -520,7 +520,7 @@ struct GameSamples *readsamples(const char **samplenames,const char *basename)
 
 	if (!i) return 0;
 
-	if ((samples = auto_malloc(sizeof(struct GameSamples) + (i-1)*sizeof(struct GameSample))) == 0)
+	if ((samples = malloc(sizeof(struct GameSamples) + (i-1)*sizeof(struct GameSample))) == 0)
 		return 0;
 
 	samples->total = i;
diff --git a/src/ost_samples.c b/src/ost_samples.c
index 553751b5..f67f0ad7 100644
--- a/src/ost_samples.c
+++ b/src/ost_samples.c
@@ -677,8 +677,9 @@ static void ost_set_last_played(int sa_left, int sa_right)
 
 static bool ost_last_played(int sa_left, int sa_right)
 {
-  if ( (last_left == sa_left) && (last_right == sa_right) )
-    return true;
+  if(sample_playing(0) == 0 || sample_playing(1) == 0)
+    if ( (last_left == sa_left) && (last_right == sa_right) )
+      return true;
 
   return false;
 }
```

grant2258:
Well there is problems in general with the ost that are probably there due to the samples not surviving a reset before to stop segfaults happening on reset. This is to fix the lifetime of the samples not to re write the ost code. It was a bug from the original core though, just an over sight auto_malloc always cleans up on reset.

mahoneyt944:
For the samples:
Malloc should be fine, though I don't see where it's freeing from at a first glance.

For the OST:
If we are going to preserve the samples on reset, I should probably create a reset hook for OST / game specific variables. Currently the OST inits these in the machine driver during install which i don't believe is called again on reset but I'd have to test and see.

grant2258:
They do free on starting and stopping the sample you can add a free on reset as well just keep in mind you hacked this in the past for journey.

mahoneyt944:
I didn't consider the samples would be continuously allocating and freeing when starting and stopping. That might be a big performance hit with how big these samples are. Are we sure we want that?

grant2258:
I dont care you can do what you like if you want to make it not work for lower end devices and always keep full ost loaded in memory all the time is your call.

mahoneyt944:
I'm not really on either side of it per say. I'm just curious if a low end device could even allocate and free these massive samples on demand without slowdowns. @KMFDManic may have experience with trying this.

grant2258:
There isint massive demands a song usually changes when during a level change it not like your flipping songs every second unless your using sloppy code to start and stop the sample all the time. Smaller samples are left loaded not sure what your actual issue is.

mahoneyt944:
@grant2258 https://github.com/libretro/mame2003-plus-libretro/commit/bac1626f65a5f695bc61587e5d441dec1e9a1001 this should fix issues with the OST on reset. Let me know if there's any bugs on reset that you noticed before this fix if still a problem.

grant2258:
@KMFDManic there is an issue with samples unloading here not sure if your port is updated. Its using channel numbers instead of a sample number but it does only load when played its just not freeing properly when it stops. Ill need to test your lates core as well when I tested shinobi on the 68k core I noticed the music wasnt playing think its ost related and might have been fixed since that was a while ago.

mahoneyt944:
Some info on this, forgot about this.
https://github.com/libretro/mame2003-plus-libretro/wiki/Building-samples-for-mame2003-plus

KMFDManic:
@mahoneyt944 @grant2258 yes forgot I had dropped that info and the source files a few years back:) 

There are definitely some issues with some of the custom samples on Xtreme that may be fixed for plus.  IE: Bad Dudes works with samples but without has no music.  Bionic Commando has muffled sound effects with samples support coded in, but music samples play fine.  Its almost surreal like how adventures of batman on robin for md/genesis has very low sound fx but clear music.

I will better outline how samples are on the Xtreme fork and am happy to test any fixes or changes.

I also never switched over to the samples outside the driver files, since it may be good to be able to test any differentiations between plus Xtreme that way.

grant2258:
@KMFDManic the mixer and other sound cores effect each other. Compare your moonwalker to ours and you'll see. I can take a look at fixing up your samples that arent mixing well but I only have access to the samples on internet archive and there aren't that many ost samples there. If i had the samples for the bad mixes I  could take a peek.

mahoneyt944:
I think everything should be at least working now. I tried tripping up the OST with resets and didn't see any crashes. I guess we could see about optimizing allocation and freeing though. If it's loading the entire sample at once maybe we could just hotswap small segments at a time or something.

grant2258:
@mahoneyt944 the only reason the samples are not freed on reset is because of journey.  I have a solution for this samples is playing should return the sample number as apposed to true or false. This is easy to implement and it wont effect any other samples. as they are just expecting a true or false. I would actually lean towards handling the samples outside of the mixer ideally but that's for another day.

KMFDManic:
@grant2258 @mahoneyt944 ill reupload any samples I have that you do not to my google drive for testing.  Thanks:)

Robocop was one of my personal favorites sample wise!  You should have fun with the shinobi samples too.  

**robocop test**

https://youtu.be/qpAGp4P_P_E?si=TK60YWhfBGuZMrtM

**shinobi test**

https://youtu.be/iNwKRsUg-xg?si=6hkRA3e88bykGpgR

KMFDManic:
@mahoneyt944 @grant2258 I am still going to do a little updating to Xtreme, as well as audit the sample sets I have added.  Kurrono initially had Bad Dudes OR Robocop working...then, later was able to have them simultaneously working on his end.  I will have to play with the modified driver file and see about adding that to Xtreme.  Where it stands, I have Robocop implemented...BUT, Bad Dudes has no sound, in my last test.  I let that go for now, since Xtreme is more so of a test Core as far as experimental stuff, whilst fixing stuff up:)

If you can spot possibly why the Deco driver doesnt play sound for Bad Dudes, while working fine for Robocop...if not being affected by the other bug/s you have brought up to me, it would be interesting to know.  

I added other custom ost additions, too, just haven't finished up test sample osts for them...Here are samples for Street Fighter 1, Robocop, Ikari Warriors, Shinobi, that all work on 2003 Xtreme.  Looking forward to any feedback, and will gladly test any trivial fix ups...I really need to get the other Custom OSTs finished, as Contra, Ghosts n Goblins, Haunted Castle, Joe and Mac, and a few others I have also added!  I did just unfinished quick test sample sets for those ones!  Let me know how these 4 fare, for now, particularly shinobi and robocop!  We can go from there.  

I can also test stuff on the end of Plus..If I recall, OutRun, I think did get screwy on "restart" of game with Xtreme.  But, it has been awhile since I tested the restart thing.  I didn't think much of it at the time!

grant2258:
@KMFDManic should be able to take a look this weekend. I have been meaning to fix the sample volumes for a while. Its not respecting the global mixing levels at all, its probably why there has been a lot of head scratching over the ost and volumes.

KMFDManic:
@grant2258 I have other drivers I can add in that may help with testing too.  BattleToads custom ost works but plays original music simultaneously at much lower volume.  Quite odd.  But yeah, the issues you refer to are likely due to all of these shenanigans.  The robocop working seemingly fine but bad dudes having no sound is probably best test for the moment. 

But can add in other drivers if it helps testing further to fix the glaring issues with volume etc

grant2258:
Im actually working on the updates now. Well the volume at least things can be backported here if they are wanted.

mahoneyt944:
There were many bugs fixed with the new OST system and it makes it much easier to troubleshoot in general. It's likely you're getting some logic crossed somewhere if you have no sound or dual sounds playing. I noticed the logic your using on most of the new implementations are structured after outrun (right down to the variables meant for outrun, like `lastwave`). You might want to audit some of those for unused variables and stuff as that might help find some of those odd issues. The old OST method works just as well as the new one....the old system is just a little more tedious to work with and easier for wires to cross.

If you wanted to update your OST, almost everything is contained in `ost_samples.c` and `ost_samples.h`. you'd just have to source clean the affected drivers (or cherry pick the drivers from 2003+). I tried to minimize the footprint within the drivers so that the OST would be more centralized.

grant2258:
@mahoneyt944 can you cherry pick this https://github.com/KMFDManic/mame2003-xtreme/pull/103/commits/40a87cecf2ca38263eeb1583760b9501c4c44dce for here when you get time.

mahoneyt944:
Sure https://github.com/libretro/mame2003-plus-libretro/commit/da2832c302bf1d4043e0077d340a8972134e12bb

grant2258:
youll need to include #include <math.h> in v60.c

mahoneyt944:
@grant2258 msvc still not compiling after including math.

mahoneyt944:
Also seems like there's some audio crackling. Try sonic, when you coin up, select sonic, and opening music.

grant2258:
for the the build fix change `truncf to` `_truncf` if fixes it ill put a fix in. Ill check on @KMFDManic  port.

grant2258:
Revert this commit and all should be fine.

mahoneyt944:
I did some googling on `truncf` and found this:

The truncf function, which truncates a float value, is not available in the C89 standard. It was introduced in the C99 standard. 

Also offered a solution, I guess we could add this when truncf is not defined
```
#include <stdio.h>

float truncf_c89(float x) {
  if (x >= 0) {
    return (float)((int)x);
  } else {
    return (float)((int)x);
  }
}
```

mahoneyt944:
@KMFDManic you could probably pull `ost_samples.c` and `ost_samples.h` then use GitHub desktop app to search 2003+ for uses within the source from that header. Other than that, you'd just have to remove all the spaghetti mess from the old system in each driver.

You might be able to shortcut some of the cleanup by dropping 2003+ drivers in directly, but you'll have to check each one for any non related changes. 

The OST portions are very simple, generally:
1. Include ost_samples.h
2. Install OST in the machine driver
3. Hook the OST routine to the sound data (like the old system)
4. If applicable hook the fade function (contra and sf2).

grant2258:
@KMFDManic you would really benefit from using the ost it works a lot better.

KMFDManic:
@grant2258 @mahoneyt944 Did preliminary separation of Custom OSTs, and results are as follows...

Contra Custom OST seems to play fine.

Ikari didn't play Custom OST, nor did RoboCop.  

Hopefully, you can see if I missed anything.  I will start getting the other OSTs over, too.

I noticed some updating was done to snk driver, and possibly by me cherry picking that exact 2003 Plus driver, I may have possibly offset or broken something?  Same with dec0 driver?

Contra works fine, however:)  And, I tested some of the remaining Custom OSTs, Shinobi in my coding works fine still, etc.

Thanks in advance for letting me know what I may have missed, in the porting, that could have potentially screwed up Ikari Warriors and/or RoboCop:)

Additionally, in ost_samples.c, I commented out:

//  if (!options.use_alt_sound) return false;
//  if (!options.content_flags[CONTENT_ALT_SOUND]) return false;

since they are nowhere in the 2003 Xtreme Coding, currently.  Hopefully, that isn't the problem:)  But, Contra does work perfectly, thus far, in testing!

https://github.com/KMFDManic/mame2003-xtreme/commit/8c1babb433b79f94364e8c23143e079f2c6fed1c

mahoneyt944:
Those are core options that disable the OST so they aren't required to work.

You'll want to call `ost_init();` from `cpu_pre_run()` in cpuexec.c

mahoneyt944:
@KMFDManic let's just focus on dec0.c driver for robocop. When you say it's not working, can you go into more detail? No sound at all? Game not booting etc.

KMFDManic:
@mahoneyt944 robocop works fine.  But custom samples dont seem to load with the ost_samples.c and h files as they should.  But, oddly, Contra does.  Likely something negligible and minor, I hope.

KMFDManic:
@mahoneyt944 robocop works fine.  But custom samples dont seem to load with the ost_samples.c and h files as they should.  But, oddly, Contra does.  Likely something negligible and minor, I hope.

mahoneyt944:
When the game is running, go into the mame menu, game information, does it say "Samples" in there?

grant2258:
@KMFDManic daft question are the samples named the same in your zip as they are in the list? I need to dl them samples better do it now.

KMFDManic:
@mahoneyt944 currently at work but will definitely check for that and any other anomalies once home later.  It's likely something trivial, as it nearly almost ends up being:) Thanks again for both you and @grant2258 pushing me to branch to this better more fully realized ost path!

KMFDManic:
@grant2258 as far as im aware the sample names in zip should match up. They worked on my previous driver fine.  I should have tested same sample zip on current 2003 plus to ensure it was just on 2003 Xtreme end.  But the robocop samples I uploaded above a few posts should load for u fine in plus.

Ikari had same behavior, normal game music but samples not seeming to load.

Now, shame I am not home to help retest stuff! But, looking forward to getting this sorted so I can add more custom osts.

mahoneyt944:
@KMFDManic I think it's because your missing the call to ost_init() from cpu_pre_run() in cpuexec.c

If you don't init this sa_volume will be 0. Contra sets this manually so likely why it worked

KMFDManic:
@mahoneyt944 feel free to pull request that change if need be.  Otherwise, I can test things once home.  Just glad you were able to quickly ascertain where the problem most likely was:)

mahoneyt944:
@KMFDManic looks like grant beat me to it. His PR should fix it.

KMFDManic:
@grant2258 @mahoneyt944 thanks a lot both of you.  Gonna focus on getting more custom osts implemented over next several days.

PR #1957, opened 2025-03-28
Update Makefile

grant2258:
error the same way msvc does on buildbot so we know about these errors when compiling locally and generally cause less issues all in.


mahoneyt944:
Might want to add to mame2003 too if it will help there too

mahoneyt944:
Also this actually fails for msvc, so you'll have to exclude it when it's msvc

grant2258:
just revert it for now ill add checks in when I get access to the main pc. There are two other fixes needed added ignore // comment error msvc allows this  and variables must be set at the start of a function.

PR #1958, opened 2025-04-02
make gcc/clang fail like mscv does

grant2258:
make clang and gcc fail like msvc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -822,11 +822,11 @@ CFLAGS += -DHAVE_SOCKLEN_T
 CFLAGS += -D_LARGEFILE_SOURCE
 CFLAGS += -D_FILE_OFFSET_BITS=64
 
-
-#make gcc fail like msvc does
+# make gcc fail like msvc does
 ifneq (,$(findstring msvc,$(platform)))
-CFLAGS += "-Werror=vla -Werror=declaration-after-statement"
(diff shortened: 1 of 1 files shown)


mahoneyt944:
I can't recall which, but there were some platforms that would not build if you remove the space from the link out line. I recall having to add it back in to correct. https://github.com/libretro/mame2003-plus-libretro/commit/670ecc1dd82052c8192704a1df1ee32700c68f30

mahoneyt944:
Still errors msvc, might need to make sure the compiler is gcc before adding the flag in.

grant2258:
its the if condition.

change
```

ifneq (,$(findstring msvc,$(platform)))
 to 
ifeq (,$(findstring msvc,$(platform)))

``` 

# make gcc fail like msvc does
 

ifneq (,$(findstring msvc,$(platform)))

grant2258:
> I can't recall which, but there were some platforms that would not build if you remove the space from the LINKOUT line. I recall having to add it back in to correct. [670ecc1](https://github.com/libretro/mame2003-plus-libretro/commit/670ecc1dd82052c8192704a1df1ee32700c68f30)

That really should be fixed properly most people set there editors to remove tailing spaces. Its easy enough to fix

PR #1964, opened 2025-04-06
fix rf5c68 and keep both compilers happy without workarounds

grant2258:
@mahoneyt944 test its fine for you then commit.

### DIFF
--- a/src/drivers/segas32.c
+++ b/src/drivers/segas32.c
@@ -2475,7 +2475,7 @@ INPUT_PORTS_END
 struct RF5C68interface sys32_rf5c68_interface =
 {
   RFC_CLOCK/4,
-  55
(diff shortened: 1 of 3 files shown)


mahoneyt944:
Ok I'll test it.

mahoneyt944:
Seems ok. I'll do more testing later though

grant2258:
sonic clips a little changing the segas32s volume fixes it if you set it to 50 rather than 55. This happens on extreme as well but not so bad sound levels are different of them is some spots due to our osd volume code being removed.

mahoneyt944:
It's much better than what we had at least. Hopefully there's a way to get it that last little bit over the line. Seems pretty sensitive to get it just right. At least it doesn't effect too many game either

grant2258:
the only complication is msvc :)

PR #1971, opened 2025-05-03
Tidy

grant2258:
@arcadez2003 libretro had a hacked mame-memory.c patch to hide a serious bug. A while ago we added something we needed to trampoline in the port read handler using static handlers like MRA_RAM MRA_BANK1 ect. I fixed it the segfault in the original code and managed to trace this and fix using static handlers in static PORT_READ(x)_START(readport ) and write handlers so they wont be an issue anymore. Ive also fixed up the memory.c common.c cheat.c 64 bit issue remaining all them errors did need looked at for a while. The fix was done a day or two ago but I had to do some testing and it all good. @mahoneyt944 I added rules for msvc to use the old style FPTR all other ports now use uintptr_t as they are not confined to c89.

### DIFF
--- a/src/cheat.c
+++ b/src/cheat.c
@@ -1991,23 +1991,23 @@ static void RebuildStringTables(void)
 		(!menuStrings.subStrings && menuStrings.numStrings) ||
 		(!menuStrings.buf && storageNeeded))
 	{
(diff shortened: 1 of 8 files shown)


grant2258:
dont pull this yet there is one test that i need to remove ill let you know when to pull

PR #1993, opened 2025-09-20
Cleanup

mahoneyt944:


grant2258:
just a heads up when using BIT always start from 0 instead of one on this core anyway not sure about later versions of mame. Only reason I mention this is because it bit me one time I assumed it started at 1. The code looks like its fine though for here just some useless information from previous escapades.

PR #2019, opened 2026-05-14
Add support for 16KB pages on Android.

Swordfish90:
Since last year Google is requiring every application on Google Play to support 16KB pages: https://android-developers.googleblog.com/2025/05/prepare-play-apps-for-devices-with-16kb-page-size.html

It's a tiny build script change, I've tested it locally and it's working fine.

We did a test pilot with fceumm (https://github.com/libretro/libretro-fceumm/pull/644) and now I'm creating all the PRs for the other cores.


mahoneyt944:
Technically the core isn't part of the play store, I'm only assuming but I'd imagine as long as the frontend "retroarch" is meeting those standards the core wouldn't matter as much. Either way that would be a decision for Libretro to have a uniform consensus on. @LibretroAdmin

LibretroAdmin:
I merged the other prs that OP sent. I see no real objection to it

PR #2024, opened 2026-08-09
Ghosts'n Goblins hacks: Enhanced Edition and Artoria Hacks + Enhanced…

GoodLuckTrying:
… Editions

Thank you wanting to make a contribution to this project!

Please note that by contributing code or other intellectual to this project you are allowing the project to make unlimited use of your contribution. As with the rest of the project, new contributions will be made available freely under the classic MAME Non-Commercial License.

**This license can be viewed at https://raw.githubusercontent.com/libretro/mame2003-plus-libretro/master/LICENSE.md**.

PR #2030, opened 2026-08-21
Ghouls 'n Ghosts - Artoria Hacks

GoodLuckTrying:
This PR is based on updates on HBMAME. Should work just fine now.

What confused me the most is how you guys handle the palette roms. Seems like for daimakai, you interleave the 4 vanilla roms into "ghl17.bin" and for the others, you just renamed "dm-17.7j" to "ghl17.bin" as well?


25minutes:
@GoodLuckTrying Has the hacks been supported? I see the pull request merged, but neither ghoulsmaiden nor ghoulsknight can work. It always complains: Game driver not found.

GoodLuckTrying:
Yeah something's weird. The hacks aren't in https://github.com/libretro/mame2003-plus-libretro/blob/master/metadata/mame2003-plus.xml

Maybe I didn't add them somewhere they also needed to be added? They're all in these drivers:
https://github.com/libretro/mame2003-plus-libretro/blob/master/src/drivers/cps1.c
https://github.com/libretro/mame2003-plus-libretro/blob/master/src/drivers/gng.c

@mahoneyt944 please show me the light!

mahoneyt944:
Replied here
https://github.com/libretro/mame2003-plus-libretro/pull/2024

25minutes:
Thanks for the update. gngmaidenenh works correctly now. 

However, it seems the driver for Ghouls 'n Ghosts - Artoria Hacks has issues. In both ghoulsmaiden and ghoulsknight, the textures are completely garbled.

GoodLuckTrying:
Alright, I dug into this one. It's not the romsets — the exact same ROMs load fine in FB Neo, so the data is good. The problem is that our new sets never get a CPS-B config.

In src/vidhrdw/cps1_vidhrdw.c, MACHINE_INIT( cps ) picks the config by doing a straight strcmp on the driver name against cps1_config_table[], and there's no fallback to the parent set:

    while(pCFG->name)
    {
        if (strcmp(pCFG->name, gamename) == 0) break;
        pCFG++;
    }
    cps1_game_config=pCFG;

    if (!cps1_game_config->name)
    {
        gamename="cps2";   /* <- everything unknown ends up here */
        ...
    }

So ghoulsmaiden, ghoulsknight, ghoulsumaiden, ghoulsuknight, daimakaimaiden and daimakaiknight all land on the generic {"cps2", NOBATTRY, 4,4,4, ...} entry instead of the CPS_B_01 row their parents use. That means wrong CPS-B registers, scroll banks 4,4,4 instead of 0,0,0, and kludge 0 instead of 1 — which is the bit that actually breaks the graphics, because the old Ghouls sprite kludge never fires:

    if (cps1_game_config->kludge == 1 && code >= 0x01000)
    {
        code += 0x4000;
    }

That also explains why the gng.c hacks came out fine and only the CPS1 ones are garbled — gng has no name-keyed table like this, so driver.c was all it needed.

Fix is just adding the six clones next to their parents, right under the daimakai row:

    {"ghoulsmaiden",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },
    {"ghoulsknight",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },
    {"ghoulsumaiden",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },
    {"ghoulsuknight",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },
    {"daimakaimaiden",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },
    {"daimakaiknight",CPS_B_01, 0,0,0, 0x0000,0xffff,0x0000,0xffff, 1 },

I'll throw a PR with the fix, if it's as I assume.

EDIT: https://github.com/libretro/mame2003-plus-libretro/pull/2037

25minutes:
Confirmed that both hacks work well now. Thanks for the awesome work!

